PR #62, opened 2025-07-16
Update the deep link in response returned by n8n_diagnostic tool

naXa777:


czlonkowski (approved):
LGTM\! The URL update correctly adds the tab parameter for proper GitHub navigation. The change from 'new-v260' to 'optional' in the anchor also makes the link more descriptive. Thanks for the fix\!

PR #67, opened 2025-07-17
Update CLAUDE_CODE_SETUP.md

Ivanfvj:
fix .mcp.json file for Claude code setup Project configuration

Claude Code expects the key `"mcpServers"` instead of `"servers"`


czlonkowski (approved):
Approved. This is a correct and necessary documentation fix. The key should indeed be 'mcpServers' not 'servers' for Claude Code configuration.

PR #151, opened 2025-08-19
Add env option for forcing IPv4.

saul-jb:
This PR adds the config option `N8N_API_FORCE_IPV4` which forces the api calls to run over IPv4.

For some ISPs/network configurations, IPv6 doesn't work reliably and node throws an error as soon as the IPv6 request fails without waiting to check if the IPv4 would succeed, so this just gives the user the option of removing IPv6 altogether if necessary.

### DIFF
--- a/src/config/n8n-api.ts
+++ b/src/config/n8n-api.ts
@@ -6,6 +6,7 @@ import { logger } from '../utils/logger';
 const n8nApiConfigSchema = z.object({
   N8N_API_URL: z.string().url().optional(),
   N8N_API_KEY: z.string().min(1).optional(),
+  N8N_API_FORCE_IPV4: z.coerce.boolean().default(false),
   N8N_API_TIMEOUT: z.coerce.number().positive().default(30000),
   N8N_API_MAX_RETRIES: z.coerce.number().positive().default(3),
 });
@@ -35,6 +36,7 @@ export function getN8nApiConfig() {
   }
   
   return {
+    ipv4: config.N8N_API_FORCE_IPV4,
     baseUrl: config.N8N_API_URL,
     apiKey: config.N8N_API_KEY,
     timeout: config.N8N_API_TIMEOUT,
--- a/src/services/n8n-api-client.ts
+++ b/src/services/n8n-api-client.ts
@@ -24,10 +24,13 @@ import {
 } from '../types/n8n-api';
 import { handleN8nApiError, logN8nError } from '../utils/n8n-errors';
 import { cleanWorkflowForCreate, cleanWorkflowForUpdate } from './n8n-validation';
+import { Agent as HttpsAgent } from 'https';
+import { Agent as HttpAgent } from 'http';
 
 export interface N8nApiClientConfig {
   baseUrl: string;
   apiKey: string;
+  ipv4: boolean;
   timeout?: number;
   maxRetries?: number;
 }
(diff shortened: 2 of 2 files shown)


czlonkowski (changes requested):
## 🔧 Changes Requested

Thanks for this useful addition! The IPv4 forcing option will help users with IPv6 connectivity issues. However, there are a few required changes before this can be merged:

### 🔴 Required Fix - Test Failure

The tests are failing because the new `ipv4` property is missing from test configurations:

**tests/unit/services/n8n-api-client.test.ts** (line 32-37) needs updating:
```typescript
const defaultConfig: N8nApiClientConfig = {
  baseUrl: 'https://n8n.example.com',
  apiKey: 'test-api-key',
  ipv4: false,  // ← Add this line
  timeout: 30000,
  maxRetries: 3,
};
```

Please check for any other test files that might need similar updates.

### 💡 Additional Improvements

1. **Optimize agent creation** - Only create agents when needed:
```typescript
const httpsAgent = ipv4 ? new HttpsAgent({ family: 4 }) : undefined;
const httpAgent = ipv4 ? new HttpAgent({ family: 4 }) : undefined;
```

2. **Add documentation** - Update `.env.example`:
```bash
# Force IPv4 for n8n API connections (useful for IPv6 issues)
# N8N_API_FORCE_IPV4=false
```

3. **Minor comment fix** - Line 64: Change "http/s agents" to "HTTP/HTTPS agents"

4. **Consider debug logging**:
```typescript
if (ipv4) {
  logger.debug('Forcing IPv4 for n8n API connections');
  config.httpAgent = httpAgent;
  config.httpsAgent = httpsAgent;
}
```

### ✅ What Works Well

- Clean implementation using Node.js built-in agents
- Proper Zod validation with sensible defaults
- Backward compatible
- Solves a real connectivity issue

Once the test file is updated (which will fix the CI failure), this will be ready to merge. The other suggestions are minor optimizations that would improve the feature but aren't blocking.

czlonkowski:
Thanks @saul-jb — forcing IPv4 for the n8n API is a legitimately useful escape hatch for broken-IPv6 environments, and it's still unimplemented in `main`. As with the self-signed-cert PR, though, this one can't merge as-is for an architectural reason.

`src/services/n8n-api-client.ts` was rewritten for SSRF protection (GHSA-cmrh-wvq6-wm9r): the request interceptor now unconditionally assigns SSRF-pinned `httpAgent`/`httpsAgent` (resolved to a validated IP) on every request. This PR attaches its own `family: 4` agents in that same interceptor, so they'd be clobbered — the feature wouldn't take effect on current `main`. The branch is also ~689 commits behind and conflicts in both touched files, and the required `ipv4` field broke existing test fixtures.

The good news is the new transport layer is a cleaner home for this: `SSRFProtection.createPinnedAgents(address, family)` already takes a `family`, and `validateWebhookUrl` resolves one — so an IPv4-force preference can be honored during resolution rather than via a separate agent. A fresh implementation against current `main` (optional `ipv4` config defaulting to false, threaded through `getN8nApiConfigFromContext`, a unit test, and an `.env.example` entry) would be the way to go.

Leaving this open pending a maintainer decision on reimplementing vs. rework. Thanks for the contribution and for flagging the problem.

czlonkowski:
Thanks. The API client now resolves DNS itself and pins each connection to the validated address with explicit IPv4/IPv6 handling (`PinnedAgents` in `src/utils/ssrf-protection.ts`, used by `src/services/n8n-api-client.ts`), which covers the IPv4-forcing case this PR added. Closing as superseded.

PR #375, opened 2025-10-26
Resolve GitHub Issue 292 in n8n-mcp

czlonkowski:
…th undefined

Add detailed documentation section for property removal pattern in n8n_update_partial_workflow tool:
- New "Removing Properties with undefined" section explaining the pattern
- Examples showing basic, nested, and batch property removal
- Migration guide for deprecated properties (continueOnFail → onError)
- Best practices for when to use undefined
- Pitfalls to avoid (null vs undefined, mutual exclusivity, etc.)

This addresses the documentation gap reported in issue #292 where users were confused about how to remove properties during node updates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,95 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [2.22.7] - 2025-10-26
+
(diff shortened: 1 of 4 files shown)


michox:
this doesn't seem to be working on VSCode Github Copilot. Claude claims that the VSCode interface won't accept undefined as a value to be passed and can't manage removing properties via the mcp. null however, would be possible.

Issue #386, opened 2025-11-01
# FTS5 Database Health Check Fails on macOS - Server Crashes Immediately


Lunifai:
## Description
The n8n-mcp server crashes immediately after initialization with an FTS5 module error during database health check. This prevents the MCP server from running at all, making it completely unusable.

## Environment
- **OS**: macOS (darwin arm64)
- **Node.js version**: v24.11.0
- **n8n-mcp installation method**: Global install via `sudo npm install -g n8n-mcp`
- **n8n instance**: Remote (Hostinger cloud hosting)
- **n8n API**: Confirmed working (tested with curl)

## Error Details

When running `n8n-mcp`, the server initializes successfully but crashes during the database health check:

```
[2025-11-01T04:57:49.040Z] [n8n-mcp] [INFO] Attempting to use better-sqlite3...
[2025-11-01T04:57:49.041Z] [n8n-mcp] [WARN] Failed to initialize better-sqlite3, falling back to sql.js
[2025-11-01T04:57:49.069Z] [n8n-mcp] [INFO] Loaded existing database from /usr/local/lib/node_modules/n8n-mcp/data/nodes.db
[2025-11-01T04:57:49.069Z] [n8n-mcp] [INFO] Successfully initialized sql.js adapter (pure JavaScript, no native dependencies)
[2025-11-01T04:57:49.076Z] [n8n-mcp] [INFO] FTS5 not available, using LIKE search for templates
[2025-11-01T04:57:49.077Z] [n8n-mcp] [INFO] Database initialized successfully from: /usr/local/lib/node_modules/n8n-mcp/data/nodes.db
[2025-11-01T04:57:49.077Z] [n8n-mcp] [INFO] MCP server initialized with 42 tools (n8n API: configured)
[2025-11-01T04:57:49.078Z] [n8n-mcp] [ERROR] Database health check failed: Error: no such module: fts5
    at e.handleError (/usr/local/lib/node_modules/n8n-mcp/node_modules/sql.js/dist/sql-wasm.js:90:371)
    at e.prepare (/usr/local/lib/node_modules/n8n-mcp/node_modules/sql.js/dist/sql-wasm.js:89:104)
    at SQLJSAdapter.prepare (/usr/local/lib/node_modules/n8n-mcp/dist/database/database-adapter.js:202:30)
[...]

PR #398, opened 2025-11-04
fix: Gracefully handle FTS5 unavailability in sql.js fallback

czlonkowski:
## Summary

Fixed critical startup crash when the server falls back to sql.js adapter due to Node.js version mismatches between build and runtime.

## Problem

When Claude Desktop uses a different Node.js version than the one used to build n8n-mcp:
- better-sqlite3 fails to load due to NODE_MODULE_VERSION mismatch (e.g., built with Node v22, running with Node v20)
- System gracefully falls back to sql.js adapter (pure JavaScript, no native dependencies)
- **BUT** the database health check crashed with "no such module: fts5" error
- Server exits immediately after startup, preventing connection

### Error Details
```
[ERROR] Database health check failed: Error: no such module: fts5
    at e.handleError (sql-wasm.js:90:371)
    at e.prepare (sql-wasm.js:89:104)
    at SQLJSAdapter.prepare (database-adapter.js:202:30)
    at N8NDocumentationMCPServer.validateDatabaseHealth (server.js:251:42)
```

**Root Cause:** The health check attempted to query the FTS5 (Full-Text Search) table, which is not available in sql.js. The error was not caught, causing the server to exit.

## Solution

Wrapped the FTS5 health check in a try-catch block to handle sql.js gracefully:

- Added try-catch around FTS5 table existence check
- Logs clear warning when FTS5 is not available
- Server continues with fallback search (LIKE queries instead of FTS5)
- Graceful degradation: works with any Node.js version

### Before Fix
- ❌ Server crashed immediately when using sql.js fallback
- ❌ Claude Desktop connection failed with Node.js version mismatches
- ❌ No way to use the MCP server without matching Node.js versions exactly

### After Fix
- ✅ Server starts successfully with sql.js fallback
- ✅ Works with any Node.js version (graceful degradation)
- ✅ Clear warning about FTS5 unavailability in logs
- ✅ Users can choose between sql.js (slower, works everywhere) or rebuilding better-sqlite3 (faster, requires matching Node version)

## Performance Notes

When using sql.js fallback:
- Full-text search (FTS5) is not available, falls back to LIKE queries
- Slightly slower search performance (~10-30ms vs ~5ms with FTS5)
- All other functionality works identically
- Database operations work correctly

**Recommendation:** For best performance, ensure better-sqlite3 loads successfully by matching Node.js versions or rebuilding:
```bash
npm rebuild better-sqlite3
```

## Files Changed

- `src/mcp/server.ts` (lines 299-317) - Added try-catch around FTS5 health check
- `package.json` - Version bump to 2.22.10
- `package.runtime.json` - Version bump to 2.22.10
- `CHANGELOG.md` - Documented fix

## Testing

- ✅ Tested with Node v20.17.0 (Claude Desktop version)
- ✅ Tested with Node v22.17.0 (build version)
- ✅ Server starts successfully in both cases
- ✅ sql.js fallback works correctly with graceful FTS5 degradation
- ✅ All 6 startup checkpoints pass
- ✅ Database health check passes with warning

## Checklist

- [x] Code compiles without errors
- [x] Tests pass
- [x] CHANGELOG.md updated
- [x] Version bumped
- [x] Commit message follows conventions
- [x] Tested with both Node versions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,102 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [2.22.10] - 2025-11-04
+
+### 🐛 Bug Fixes
+
+**sql.js Fallback: Fixed Database Health Check Crash**
+
+Fixed critical startup crash when the server falls back to sql.js adapter (used when better-sqlite3 fails to load, such as Node.js version mismatches between build and runtime).
+
+#### Problem
+
+When Claude Desktop was configured to use a different Node.js version than the one used to build the project:
+- better-sqlite3 fails to load due to NODE_MODULE_VERSION mismatch (e.g., built with Node v22, running with Node v20)
+- System gracefully falls back to sql.js adapter (pure JavaScript, no native dependencies)
+- **BUT** the database health check crashed with "no such module: fts5" error
+- Server exits immediately after startup, preventing connection
+
+**Error Details:**
+```
+[ERROR] Database health check failed: Error: no such module: fts5
+    at e.handleError (sql-wasm.js:90:371)
+    at e.prepare (sql-wasm.js:89:104)
+    at SQLJSAdapter.prepare (database-adapter.js:202:30)
+    at N8NDocumentationMCPServer.validateDatabaseHealth (server.js:251:42)
+```
+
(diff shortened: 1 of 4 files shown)


czlonkowski:
This PR fixes issue #386 - FTS5 Database Health Check crash on macOS when sql.js fallback is used.

PR #459, opened 2025-12-01
Fix Docker healthcheck

tolgatuncoglu:
This PR fixes Docker healthcheck behavior for the n8n-mcp container.

n8n-mcp runs as a stdio-based MCP server (node dist/mcp/index.js) and does not expose an HTTP server on port 3000. A curl-based healthcheck to http://127.0.0.1:3000 always fails and causes Docker to mark the container as unhealthy even though the MCP server is working correctly.

Changes:
- Replace the curl-based healthcheck in docker-compose.n8n.yml with a simple process-based healthcheck
- Add a short note in the README explaining that n8n-mcp is not an HTTP server

### DIFF
--- a/README.md
+++ b/README.md
@@ -482,6 +482,8 @@ Want to use n8n-MCP with your n8n instance? Check out our comprehensive [n8n Dep
 - Cloud deployment on Hetzner, AWS, and other providers
 - Troubleshooting and security best practices
 
+> **Note on Docker healthchecks**: When using `docker-compose.n8n.yml`, n8n-mcp runs as a stdio-based MCP server (Node process) and does not expose an HTTP server on port 3000. Avoid curl-based healthchecks such as `curl 127.0.0.1:3000`, as they will always fail. The provided docker-compose file uses a process-based healthcheck that verifies the MCP Node process is running.
+
 ## 💻 Connect your IDE
 
 n8n-MCP works with multiple AI-powered IDEs and tools. Choose your preferred development environment:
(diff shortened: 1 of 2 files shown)


czlonkowski:
## PR Review Summary

Thank you for this contribution and for thinking about Docker healthcheck reliability!

However, after thorough review by both code and deployment reviewers, we found a **critical factual issue** with the premise of this PR.

### Finding: docker-compose.n8n.yml uses HTTP mode

The `docker-compose.n8n.yml` file explicitly sets `MCP_MODE=http` (line 44):

```yaml
environment:
  - MCP_MODE=http          # HTTP mode, NOT stdio
  - PORT=${MCP_PORT:-3000}
```

This means:
1. The container **does** run an HTTP server on port 3000
2. The `/health` endpoint **is** available and returns detailed status
3. The existing curl-based healthcheck **is correct** for this configuration

### Why the proposed change would be a downgrade

| Aspect | Current (curl) | Proposed (ps grep) |
|--------|---------------|-------------------|
| Detects hung processes | ✅ Yes | ❌ No |
| Returns health metrics | ✅ Yes | ❌ No |
| Kubernetes-compatible | ✅ Full | ⚠️ Limited |
| Application-level health | ✅ Yes | ❌ No |

### Recommendation

If you were experiencing healthcheck failures, the root cause is likely one of:
- Container startup timing (try increasing `start_period`)
- Port configuration issues
- Environment variables not being passed correctly

We'd be happy to help debug the actual issue if you can share more details about what errors you were seeing.

---

*Review conducted by code-reviewer and deployment-engineer agents*

czlonkowski:
Closing this PR as the existing curl-based healthcheck is correct for the HTTP mode configuration in docker-compose.n8n.yml. If you're experiencing specific healthcheck failures, please open an issue with details and we'll investigate the root cause.

Issue #376, opened 2025-10-26
JSON parsing errors in Claude Desktop - "Unexpected token" errors


barszczxl2:
## Problem Description
When using n8n-mcp with Claude Desktop, I'm getting multiple JSON parsing errors in the UI. However, the MCP server functions correctly despite these errors.

## Environment
- **OS:** macOS
- **Claude Desktop version:** Latest
- **n8n-mcp version:** 2.22.7
- **n8n instance:** https://localhost:5678
- **Node version:** v24.4.0

## Error Messages
Multiple errors appear in Claude Desktop UI:
- `Unexpected token '"', " "po"... is not valid JSON`
- `Unexpected token '"', " 0,\n' +"... is not valid JSON`
- `Unexpected token '"', " '},\n' +"... is not valid JSON`
- `Unexpected token '"', " "pa"... is not valid JSON`
- And many more similar errors

## Configuration
My `claude_desktop_config.json`:
```json
{
  "mcpServers": {
    "n8n": {
      "command": "npx",
      "args": ["-y", "n8n-mcp"],
      "env": {
        "N8N_API_URL": "https://localhost:5678",
        "N8N_API_KEY": "***",
        "NODE_TLS_REJECT_UNAUTHORIZED": "0"
      }
    }
  }
}
```

## What Works
Despite the errors:
- ✅ MCP connection establishes successfully
- ✅ `n8n_health_check` returns status: "ok"
- ✅ All 38 tools are available
- ✅ Workflows can be created and listed
- ✅ Basic operations work correctly

## Expected Behavior
No JSON parsing errors should appear in Claude Desktop UI when MCP operations succeed.

## Additional Context
This appears to be related to the MCP server outputting non-JSON text to stdout, contaminating the JSON-RPC communication channel. Similar issues have been reported in other MCP servers:
- https://github.com/yamadashy/repomix/issues/440
- https://github.com/modelcontextprotocol/servers/issues/133

## Screenshots
[Attach your screenshots showing the errors]

## Possible Solution
The MCP server should ensure that only valid JSON-RPC 2.0 messages are sent to stdout, with all logging/diagnostic output redirected to stderr.

DanielIshi:
## ✅ Solution Found: Add  to Configuration

I encountered the exact same issue and found the solution. The problem is that n8n-mcp outputs debug logs to STDOUT, which Claude Desktop tries to parse as JSON-RPC messages.

### Fix: Add  Environment Variable

Update your :

```json
[...]

PR #493, opened 2025-12-16
Fix n8n_update_partial_workflow tool to comply with MCP validation requirements

medinavs:
This PR fixes MCP tool schema validation for the n8n_update_partial_workflow tool.

The n8n_update_partial_workflow tool was failing MCP schema validation with error: "400 Invalid schema for function 'n8n_update_partial_workflow': In context=('properties', 'operations', 'items'), 'additionalProperties' is required to be supplied and to be false."

The "items" param from "operations" object had ``"additionalProperties" : true` which violated MCP's strict schema requirements and caused the tool to be rejected during validation.

Changes:

-Changed additionalProperties from true to false in the items schema definition of the operations object

### DIFF
--- a/src/mcp/tools-n8n-manager.ts
+++ b/src/mcp/tools-n8n-manager.ts
@@ -138,7 +138,7 @@ export const n8nManagementTools: ToolDefinition[] = [
           description: 'Array of diff operations to apply. Each operation must have a "type" field and relevant properties for that operation type.',
           items: {
             type: 'object',
-            additionalProperties: true
+            additionalProperties: false
           }
         },
         validateOnly: {


czlonkowski:
## Review Summary

Thank you for identifying this schema validation issue and taking the time to submit a PR! We appreciate the contribution.

However, after thorough investigation, we've determined that this is **not a bug in n8n-mcp** but rather an **OpenAI-specific requirement** that doesn't apply to our primary target environment (Claude Code/Desktop).

### Findings

1. **The current implementation works correctly** - We tested `n8n_update_partial_workflow` with both simple and complex operations in Claude Code, and all calls succeed without schema validation errors.

2. **The schema design is intentional** - The `additionalProperties: true` setting exists by design (see line 130 comment: "Allow any extra properties Claude Desktop might add").

3. **OpenAI vs MCP specification** - The error you encountered (`additionalProperties is required to be supplied and to be false`) is an [OpenAI API-specific requirement](https://github.com/github/github-mcp-server/issues/376), not part of the MCP specification. Claude Code/Desktop does not enforce this restriction.

4. **The proposed fix would break functionality** - Changing to `additionalProperties: false` without defining all 16+ operation type properties would cause ALL operation calls to be rejected (since `type`, `nodeId`, `node`, etc. would become "additional properties").

### If you need OpenAI compatibility

If you're using an OpenAI-based MCP client, the proper solution would be a schema transformation layer that converts schemas for OpenAI compatibility, rather than changing the core schema which would break Claude Desktop compatibility.

### References

- [OpenAI API Compatibility Issue - github-mcp-server #376](https://github.com/github/github-mcp-server/issues/376)
- [OpenAI function tool schema validation - camel-ai #2231](https://github.com/camel-ai/camel/issues/2231)
- [MCP Tools Specification](https://modelcontextprotocol.io/specification/draft/server/tools)

Thank you again for the contribution! If you have questions or believe we've missed something, please let us know.

czlonkowski:
Closing as this is an OpenAI-specific requirement, not an MCP specification issue. The current implementation is correct for Claude Code/Desktop. See the detailed explanation above.

PR #593, opened 2026-02-12
fix: add missing object unary operators

### LINKED ISSUE #592: Auto-sanitization breaks all `object` type IF operators
## Summary

The auto-sanitization system incorrectly classifies `empty`, `notEmpty`, `exists`, and `notExists` operations on `object` type as binary operators, stripping the required `singleValue: true` property. This silently breaks IF nodes that check whether an object is empty/non-empty or exists, causing them to error on execution.

## Steps to reproduce

1. Create a workflow with an IF node using an `object`/`empty` operator:

    ```json
    {
      "operator": {
        "type": "object",
        "operation": "empty",
        "singleValue": true
      }
    }
    ```

2. Make any unrelated change to the workflow via `n8n_update_partial_workflow` (e.g. rename a different node, update a connection, change settings).

3. Observe that the IF node's operator has been modified to:

    ```json
    {
      "operator": {
        "type": "object",
        "operation": "empty"
      }
    }
    ```

4. Execute the workflow. The IF node now fails with: `Wrong type: '' is a string but was expecting an object`

## Expected behavior

`singleValue: true` should be preserved (or added) for all `object` type operators, since they are all unary and do not take a right-hand value.

## Actual behavior

- `n8n_update_partial_workflow`: Silently strips `singleValue: true` from ALL `object` operators in the workflow, not just modified nodes.
- `n8n_update_full_workflow`: Rejects the entire update with an explicit error:

  > `binary operator "empty" should not have "singleValue: true". Only unary operators (isEmpty, isNotEmpty, true, false, isNumeric) need this property.`

## Root cause

`isUnaryOperator()` in `node-sanitizer.ts` and the unary list in `validateOperatorStructure()` in `n8n-validation.ts` only recognize `isEmpty`, `isNotEmpty`, `true`, `false`, `isNumeric`.

The `object` data type uses different operation names (`empty`, `notEmpty`, `exists`, `notExists`) which are all unary but missing from these lists. Per n8n source ([`packages/frontend/editor-ui/src/features/ndv/parameters/components/FilterConditions/constants.ts`](https://github.com/n8n-io/n8n/blob/4a2c06ea54c9501ea242866e78cf38dcb692b735/packages/frontend/editor-ui/src/features/ndv/parameters/components/FilterConditions/constants.ts#L252-L275)), all 4 object operators have `singleValue: true`.

## Workaround

Use the n8n REST API directly via HTTP (bypassing the MCP tool) for workflow write operations.

## Environment

- n8n-mcp version: 2.35.2
- n8n IF node typeVersion: 2.2 and 2.3

geraldk1:
The isUnaryOperator() function and validateOperatorStructure() only recognized string/boolean unary operators (isEmpty, etc.) but not object type operators (empty, notEmpty, exists, notExists). This caused sanitizeOperator() to strip the required singleValue:true from object type IF conditions, silently breaking them on execution.

fixes #592

### DIFF
--- a/src/services/n8n-validation.ts
+++ b/src/services/n8n-validation.ts
@@ -601,7 +601,7 @@ export function validateOperatorStructure(operator: any, path: string): string[]
 
   // Check singleValue based on operator type
   if (operator.operation) {
-    const unaryOperators = ['isEmpty', 'isNotEmpty', 'true', 'false', 'isNumeric'];
+    const unaryOperators = ['isEmpty', 'isNotEmpty', 'empty', 'notEmpty', 'exists', 'notExists', 'true', 'false', 'isNumeric'];
     const isUnary = unaryOperators.includes(operator.operation);
 
     if (isUnary) {
--- a/src/services/node-sanitizer.ts
+++ b/src/services/node-sanitizer.ts
@@ -237,6 +237,10 @@ function isUnaryOperator(operation: string): boolean {
   const unaryOps = [
     'isEmpty',
     'isNotEmpty',
+    'empty',
+    'notEmpty',
+    'exists',
+    'notExists',
     'true',
     'false',
     'isNumeric'
--- a/tests/unit/services/node-sanitizer.test.ts
+++ b/tests/unit/services/node-sanitizer.test.ts
@@ -220,6 +220,71 @@ describe('Node Sanitizer', () => {
       expect(sanitized).toEqual(node);
     });
 
+    it('should add singleValue for object type unary operators', () => {
+      const node: WorkflowNode = {
+        id: 'test-if-object-unary',
+        name: 'IF Object Empty',
+        type: 'n8n-nodes-base.if',
+        typeVersion: 2.2,
+        position: [0, 0],
+        parameters: {
+          conditions: {
+            conditions: [
+              {
+                id: 'condition1',
+                leftValue: '={{ $json.value }}',
+                rightValue: '',
+                operator: {
+                  type: 'object',
+                  operation: 'empty'
+                  // Missing singleValue
+                }
+              }
+            ]
+          }
+        }
+      };
+
+      const sanitized = sanitizeNode(node);
+      const condition = (sanitized.parameters.conditions as any).conditions[0];
+
+      expect(condition.operator.singleValue).toBe(true);
+    });
+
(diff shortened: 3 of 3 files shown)


PhilflowIO:
Hey @geraldk1, thanks for working on this — we're fixing the same issue (#592) in #596.

One important note: your PR description says the old operators (`isEmpty`, `isNotEmpty`) are "string/boolean unary operators" and the new ones (`empty`, `notEmpty`, `exists`, `notExists`) are "object type operators". That's not quite right.

`isEmpty` and `isNotEmpty` **do exist** in n8n, but only in the legacy **IF V1** node. The sanitizer never processes V1 nodes — it has an explicit version gate:

```typescript
function isFilterBasedNode(nodeType: string, typeVersion: number): boolean {
  if (nodeType === 'n8n-nodes-base.if') {
    return typeVersion >= 2.2; // V1 is skipped entirely
  }
  ...
}
```

| | IF V1 (typeVersion 1) | IF V2.2+ (typeVersion ≥ 2.2) |
|---|---|---|
| **Operator names** | `isEmpty`, `isNotEmpty` | `empty`, `notEmpty`, `exists`, `notExists` |
| **Parameter structure** | `conditions.string[].operation` (flat string) | `conditions.conditions[].operator.operation` (object with `singleValue`) |
| **Processed by sanitizer?** | **No** | **Yes** |

So the V1 names (`isEmpty`, `isNotEmpty`, `isNumeric`) can never reach `isUnaryOperator()` — they're dead code in that list. All 6 types in V2+ (string, number, boolean, dateTime, array, object) use the same four unary operators: `exists`, `notExists`, `empty`, `notEmpty`. You can verify this in the [source of truth — `OPERATORS_BY_ID` in constants.ts](https://github.com/n8n-io/n8n/blob/4a2c06ea54c9501ea242866e78cf38dcb692b735/packages/frontend/editor-ui/src/features/ndv/parameters/components/FilterConditions/constants.ts#L6-L275).

Your fix works functionally (the correct operators get recognized), but the old V1 names left in the lists are dead code that can never match. Our PR #596 replaces them entirely and also fixes `inferDataType()` and the docs/guidance messages.

geraldk1:
@PhilflowIO Thanks for the follow-up. Closing in favor of #596

PR #595, opened 2026-02-13
fix(mcp-http): keep streamable sessions alive on GET

yamachu:
## Summary
- Keep StreamableHTTP sessions alive by refreshing access time on GET /mcp when a session ID is provided.
- Return 404 for missing StreamableHTTP sessions to avoid SSE fallback and session churn.
- Reduces repeated reconnect loops observed with MCP client tools.

## Reproduction
- Start the MCP server in HTTP (streamable) mode.
- Connect from an MCP client(e.g. n8n AI Agent MCP Client) that uses StreamableHTTP.
- Wait 5–10 minutes to confirm the session does not churn.
- Observe repeated GET /mcp calls and frequent session churn.

### DIFF
--- a/src/http-server-single-session.ts
+++ b/src/http-server-single-session.ts
@@ -1065,6 +1065,9 @@ export class SingleSessionHTTPServer {
       const sessionId = req.headers['mcp-session-id'] as string | undefined;
       const existingTransport = sessionId ? this.transports[sessionId] : undefined;
       if (existingTransport && existingTransport instanceof StreamableHTTPServerTransport) {
+        // Refresh lastAccess so the idle-session reaper does not prune a client
+        // that is holding a long-lived GET stream without POST.
+        this.updateSessionAccess(sessionId!);
         // Let the StreamableHTTPServerTransport handle the GET request
         try {
           await existingTransport.handleRequest(req, res, undefined);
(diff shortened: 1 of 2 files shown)


czlonkowski:
Thanks @yamachu — this is a real bug and the fix is the right idea. Verified against current `main`: the `GET /mcp` handler still does **not** call `updateSessionAccess` (only the POST path does), while the idle-session reaper (`SESSION_TIMEOUT_MINUTES`, default 30m) actively prunes by `lastAccess` — so a client holding a long-lived StreamableHTTP GET stream without POSTing can indeed get reaped, causing the reconnect loops you're seeing. The keepalive is genuinely useful and not yet in main.

It can't merge as-is, though — the branch is ~165 commits behind and the `GET /mcp` hunk conflicts hard. The handler was refactored (shared `authenticateRequest`, an `existingTransport instanceof StreamableHTTPServerTransport` guard, an SSE-redirect fallback, and an `N8N_MODE` branch), so the `if (transport) {...} else { 404 }` shape no longer applies. Could you:

1. **Rebase onto current `main`.**
2. **Re-express the keepalive** by adding `this.updateSessionAccess(sessionId)` inside the existing `if (existingTransport && existingTransport instanceof StreamableHTTPServerTransport)` block, before `handleRequest`. That's the high-value, low-risk part.
3. **Reconcile the 404-on-unknown-session** behavior with main's newer fallbacks — a stale `mcp-session-id` should still be able to reach the SSE-redirect / `N8N_MODE` paths, so the 404 should only fire for non-SSE, non-n8n GETs (or be dropped if it interferes).
4. **Add a unit test** in `tests/unit/http-server-session-management.test.ts` asserting GET refreshes `lastAccess` and that an unknown session returns the intended status.

Happy to re-review once it's rebased — the keepalive line alone would be an easy merge.

yamachu:
Thank you for the detailed review!

Rather than rebasing the original branch, I created a new branch from the current main and re-expressed the changes there.

One note on the tests: asserting that `lastAccess` is refreshed required `Object.setPrototypeOf` to make the injected transport pass the `instanceof StreamableHTTPServerTransport` check. The existing mock uses `vi.fn() + return plainObject`, which breaks `instanceof`, so this was necessary to enter the keepalive branch. It's a bit of a workaround, but it's isolated to that one test.

PR #607, opened 2026-02-20
fix: preprocess JSON string params in Zod schemas for MCP compatibility

Yanndotai:
## Summary
- Some MCP clients (e.g. Claude Code) serialize array and object tool parameters as JSON strings instead of native JavaScript types
- This causes Zod validation to fail with `Expected array, received string` on `n8n_update_partial_workflow`, `n8n_update_full_workflow`, and `n8n_create_workflow`
- Added `z.preprocess()` wrappers that transparently parse JSON strings back to arrays/objects before validation, while leaving native types untouched

## Affected schemas
- `workflowDiffSchema` → `operations` field
- `createWorkflowSchema` → `nodes`, `connections` fields
- `updateWorkflowSchema` → `nodes`, `connections`, `settings` fields

## How it works
```typescript
const jsonArrayPreprocess = (val: unknown): unknown => {
  if (typeof val === 'string') {
    try { const parsed = JSON.parse(val); if (Array.isArray(parsed)) return parsed; } catch {}
  }
  return val;
};

// Before: fails when MCP client sends string
operations: z.array(z.object({...}))

// After: transparently handles both string and native array
operations: z.preprocess(jsonArrayPreprocess, z.array(z.object({...})))
```

## Test plan
- [x] TypeScript build passes with no errors
- [x] Tested with Claude Code MCP client: `n8n_update_partial_workflow` now accepts `operations` as JSON string
- [x] Tested with Claude Code MCP client: `n8n_update_full_workflow` now accepts `nodes` and `connections` as JSON strings
- [x] Native array/object parameters continue to work unchanged (preprocess is a no-op for non-string values)


diborque2733:
Hi @Yanndotai, thanks for opening this. Heads-up that the underlying compatibility issue this PR was addressing is now fixed on `main` via #727 / v2.47.8 (merged 2026-04-14):

> **fix: parse JSON-stringified workflow params in MCP schemas** — Wraps `nodes`, `connections`, and `settings` in `createWorkflowSchema` and `updateWorkflowSchema` (and `tags` in `listWorkflowsSchema`) with `z.preprocess(tryParseJson, ...)`. This mirrors the prior fix for `n8n_update_partial_workflow`'s `operations` parameter from #611/#600.

Combined with the earlier `tryParseJson` rollout for `n8n_update_partial_workflow`, the `JSON-string-instead-of-array` serialization from VS Code + GitHub Copilot and similar MCP clients is now handled across all affected workflow endpoints.

The branch here has also gone into `CONFLICTING` state against `main` after the v2.47.x series of validator and schema changes.

If v2.47.8 does not cover a case you still see broken, please leave a minimal repro and I'll take another look — but otherwise this PR can likely be closed as superseded. I'll defer closing to you in case there's context I'm missing.

PR #612, opened 2026-03-01
fix: add explicit radix to all parseInt calls

ashsolei:
## Summary

Add explicit radix parameter (`10`) to all `parseInt()` calls that were missing it across the codebase.

## Problem

Without an explicit radix, `parseInt` defaults to base 10 for decimal strings but can misparse strings with leading zeros as octal in older engines. Missing the radix parameter also violates ESLint's `radix` rule.

## Changes

16 `parseInt()` calls updated across 8 source files:

| File | Occurrences | Context |
|------|-------------|---------|
| `src/utils/n8n-errors.ts` | 1 | `retry-after` header parsing |
| `src/mcp/server.ts` | 1 | Array index notation parsing |
| `src/scripts/fetch-templates.ts` | 2 | `METADATA_LIMIT` and `OPENAI_BATCH_SIZE` env vars |
| `src/http-server.ts` | 2 | `PORT` env var |
| `src/http-server-single-session.ts` | 3 | `PORT`, `AUTH_RATE_LIMIT_WINDOW`, `AUTH_RATE_LIMIT_MAX` |
| `src/services/node-documentation-service.ts` | 1 | Version number extraction |
| `src/templates/batch-processor.ts` | 1 | Template ID parsing |
| `src/templates/metadata-generator.ts` | 3 | Template ID parsing |

Note: `http-server-single-session.ts:113` (sessionTimeout) already had the radix parameter and was not modified.

## Testing

Pure mechanical change — adds `, 10` to each call. No behavioral difference for valid decimal inputs. Prevents potential misparse of zero-prefixed strings.

### DIFF
--- a/src/http-server-single-session.ts
+++ b/src/http-server-single-session.ts
@@ -837,10 +837,7 @@ export class SingleSessionHTTPServer {
     
     // Root endpoint with API information
     app.get('/', (req, res) => {
-      const port = parseInt(process.env.PORT || '3000');
-      const host = process.env.HOST || '0.0.0.0';
-      const baseUrl = detectBaseUrl(req, host, port);
-      const endpoints = formatEndpointUrls(baseUrl);
+      const port = parseInt(process.env.PORT || '3000', 10);
       
       res.json({
         name: 'n8n Documentation MCP Server',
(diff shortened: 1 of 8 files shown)


ashsolei:

PR #613, opened 2026-03-01
fix: remove debug console.logs and use logger in library modules

ashsolei:
## Summary

Remove leftover debug `console.log` statements and replace bare `console.log`/`warn`/`error` calls with the project's `logger` utility in library modules.

## Problem

- `config-validator.ts` contained 3 leftover `console.log('DEBUG: ...')` statements inside a conditional block. In stdio mode, any direct `console.log` output corrupts the JSON-RPC transport stream.
- `docs-mapper.ts` and `node-loader.ts` used bare `console.log`/`warn`/`error` for operational logging instead of the project's `logger` utility, which properly suppresses output in stdio mode via the `MCP_MODE` environment variable check.

## Changes

| File | Change |
|------|--------|
| `src/services/config-validator.ts` | Removed 3 `console.log('DEBUG: ...')` statements and their containing conditional block |
| `src/mappers/docs-mapper.ts` | Replaced 4 `console.log` calls with `logger.warn`/`logger.debug` |
| `src/loaders/node-loader.ts` | Replaced 7 `console.log`/`warn`/`error` calls with `logger.info`/`debug`/`warn`/`error` |

## Testing

- The `logger` utility is already well-tested and used throughout the codebase (18+ existing imports)
- In stdio mode: logger suppresses all output, preventing JSON-RPC corruption
- In HTTP mode: logger suppresses output during active MCP requests
- In rebuild/validate scripts: logger outputs normally since `MCP_MODE` is not set

### DIFF
--- a/src/loaders/node-loader.ts
+++ b/src/loaders/node-loader.ts
@@ -1,4 +1,5 @@
 import path from 'path';
+import { logger } from '../utils/logger';
 
 export interface LoadedNode {
   packageName: string;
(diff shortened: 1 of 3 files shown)


ashsolei:

PR #614, opened 2026-03-01
fix: add null safety guards in config-manager and cache-utils

ashsolei:
Two null safety fixes:

**config-manager.ts**: After JSON.parse, the parsed result could be `null` (valid JSON). The code immediately accesses `this.config.userId` via non-null assertion, which would throw a TypeError. Added a guard to validate the parsed value is a non-null object before assignment, falling through to defaults on invalid config.

**cache-utils.ts**: `lastError` in `withRetry()` is declared but never initialized. If called with `maxAttempts: 0`, the for-loop never executes and `lastError.message` accesses undefined. Initialized with a sentinel Error for a meaningful message.

### DIFF
--- a/src/telemetry/config-manager.ts
+++ b/src/telemetry/config-manager.ts
@@ -200,7 +200,14 @@ export class TelemetryConfigManager {
 
     try {
       const rawConfig = readFileSync(this.configPath, 'utf-8');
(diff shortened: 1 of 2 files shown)


ashsolei:

PR #619, opened 2026-03-06
feat: add support for cookie-based authentication in n8n API client

aashsach:
- Introduced 'x-n8n-cookie' header for browser-session-based authentication.
- Updated instance context and API client to handle both API key and cookie.
- Enhanced validation for cookie format and added tests for cookie-based sessions.
- Updated documentation to reflect new authentication options and requirements.

This change allows users to authenticate using a session cookie as an alternative to the API key, improving flexibility for different deployment scenarios.

### DIFF
--- a/src/config/n8n-api.ts
+++ b/src/config/n8n-api.ts
@@ -6,6 +6,8 @@ import { logger } from '../utils/logger';
 const n8nApiConfigSchema = z.object({
   N8N_API_URL: z.string().url().optional(),
   N8N_API_KEY: z.string().min(1).optional(),
+  /** Session cookie for browser-session-based authentication (alternative to N8N_API_KEY) */
+  N8N_API_COOKIE: z.string().min(1).optional(),
   N8N_API_TIMEOUT: z.coerce.number().positive().default(30000),
   N8N_API_MAX_RETRIES: z.coerce.number().positive().default(3),
 });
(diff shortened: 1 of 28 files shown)


czlonkowski:
Thanks @aashsach — this is a clean, well-tested, additive feature. I traced the full credential flow (`config` → `getN8nApiConfigFromContext` → `new N8nApiClient(config)`) and the cookie threads through correctly; API-key precedence, cookie-only, no-creds, persistence round-trip, and placeholder rejection are all covered. The multi-tenant cache key correctly includes the cookie, so client isolation is preserved.

Three things gate merge:

**Blocking**
1. **Secret-redaction gap.** The new `x-n8n-cookie` header is not in `SENSITIVE_HEADER_NAMES` (`src/utils/redaction.ts`), but `redactHeaders()` runs on full request headers for debug logging. So a tenant's session cookie would be logged in cleartext — a secret-disclosure regression for the very feature being added. Please add `'x-n8n-cookie'` to that set and a test asserting it's redacted.
2. **Rebase.** The branch is CONFLICTING/DIRTY against current `main` — mostly diagnostic/error-text drift (e.g. the `N8N_API_KEY` message and its matching test expectations). Rebase, reconcile the strings so code and tests stay in lockstep, and re-run `npm run build && npm test`.

**Maintainer question (for @czlonkowski):** the open product question is whether an n8n **UI session cookie** actually authenticates against the public `/api/v1` REST API, or only the internal `/rest` UI endpoints. The public API is documented as `X-N8N-API-KEY`-only; if `/api/v1` rejects cookies, this path won't work end-to-end even though the code is correct. Worth confirming against a live instance before relying on it.

**Nice to have:** add CRLF/control-char validation in `isValidCookie` (defends against header injection from untrusted multi-tenant input), and a test for the `x-n8n-cookie` → `InstanceContext` inbound mapping.

The code is solid — once the redaction fix + rebase land and the `/api/v1` cookie question is confirmed, this is a good merge.

PR #632, opened 2026-03-13
chore: refresh lockfile dependencies

rotprods:
Refreshes package-lock.json based on current dependency graph.


czlonkowski:
Closing — stale lockfile-only refresh with no functional changes.

PR #645, opened 2026-03-18
fix: implement transferWorkflow operation in n8n_update_partial_workflow (#644)

### LINKED ISSUE #644: Feature Request: Add projectId support to workflow create/update tools
### Summary

Add `projectId` parameter support to workflow management tools so workflows can be created in (and moved to) specific team projects.

### Current Behavior

- `n8n_create_workflow`: No `projectId` parameter documented → new workflows always land in the personal default project
- `n8n_update_full_workflow` / `n8n_update_partial_workflow`: No way to move an existing workflow to a different project

### Desired Behavior

1. **`n8n_create_workflow`**: Accept an optional `projectId` parameter so the workflow is created directly in the correct team project.

2. **`n8n_update_partial_workflow`**: Add a new metadata operation (e.g. `transferProject`) or expose `projectId` in `updateSettings` so an existing workflow can be moved to a different project — equivalent to the n8n UI "Move to project" action or the REST API `PUT /workflows/{id}/transfer` endpoint.

### Why This Matters

Teams using n8n in a multi-project setup (personal vs. team projects) currently have no way to place or move workflows into the correct project via MCP tools. Every new workflow created through Claude/MCP lands in the personal default project and must be manually moved via the n8n UI.

### Suggested Implementation

- `n8n_create_workflow`: pass `projectId` to the underlying `POST /workflows` API call (the REST API already supports this)
- `n8n_update_partial_workflow`: add a `transferProject` operation type that calls `PUT /workflows/{id}/transfer` with `{ destinationProjectId }` under the hood

### Context

Discovered while building finance automation workflows with Claude Code + n8n-mcp. The `POST /data-tables` endpoint has a similar limitation (ignores `projectId` in the body), but that is a known n8n API bug — the workflow case is solvable on the MCP layer.

djakielski:
## Summary

- Implements the `transferWorkflow` diff operation that was defined in types but missing from the engine's switch statements
- Adds `validateTransferWorkflow()` — validates `destinationProjectId` is present
- Adds `applyTransferWorkflow()` — sets the `transferToProjectId` flag in the diff result
- After workflow update: calls `client.transferWorkflow()` when `diffResult.transferToProjectId` is set

## Root Cause

The `TransferWorkflowOperation` type, the `transferToProjectId` tracking field, and the API client method (`PUT /workflows/{id}/transfer`) were all already present. Only the `case 'transferWorkflow'` entries were missing in `validateOperation()` and `applyOperation()`, causing the operation to fall into the `default` branch with "Unknown operation type".

## Test plan

- [x] `validateOperation` returns `null` for valid `transferWorkflow` operation
- [x] `validateOperation` returns error when `destinationProjectId` is missing
- [x] `applyDiff` result includes `transferToProjectId` when operation is present
- [x] `client.transferWorkflow()` is called after workflow update
- [x] All 2516 existing unit tests pass

Fixes #644


czlonkowski:
Thank you for this contribution, @djakielski! The implementation was well done — the types, API client method, and handler logic were all solid.

We've cherry-picked your work into PR #649 with the following improvements:
- Zod schema uses `.min(1)` to reject empty strings at the input layer
- Validator signature consistent with all other validators
- `transferToProjectId` uses `|| undefined` pattern matching existing fields
- `inferIntentFromOperations` includes a `transferWorkflow` case
- Tool documentation fully updated (description, examples, use cases, pitfalls)
- Additional test scenarios (ordering, atomic mode, edge cases)
- API client tests use `simulateError` pattern matching existing tests
- Fixed two pre-existing bugs found during review

Your contribution is credited via `Co-Authored-By` in the commit.

Closing in favor of #649.

PR #658, opened 2026-03-23
feat(auth): add OAuth 2.1 support with n8n as authorization server

rneuplanche:
## Summary
- Add OAuth 2.1 authentication mode (`AUTH_MODE=oauth|both|token`) with n8n as the authorization server
- Fix missing `expiresAt` in `N8nTokenVerifier` AuthInfo, which caused MCP SDK's `requireBearerAuth` to reject verified tokens with a 500 error
- Merge header and OAuth instance contexts so `x-n8n-key` header works alongside OAuth-derived n8n URL, enabling management tools (list/create/update workflows) in OAuth mode

## Test plan
- [ ] `npm test` passes
- [ ] Server starts in `AUTH_MODE=oauth` with `OAUTH_ISSUER_URL` pointing to n8n
- [ ] Claude Code `/mcp` reconnect succeeds after OAuth flow
- [ ] Management tools (n8n_list_workflows, etc.) appear and work with `x-n8n-key` header

### DIFF
--- a/src/auth/n8n-token-verifier.ts
+++ b/src/auth/n8n-token-verifier.ts
@@ -1,0 +1,62 @@
+import type { OAuthTokenVerifier } from '@modelcontextprotocol/sdk/server/auth/provider.js';
+import type { AuthInfo } from '@modelcontextprotocol/sdk/server/auth/types.js';
+import { logger } from '../utils/logger';
+
+/**
+ * Verifies OAuth access tokens issued by n8n's authorization server.
+ * n8n's OAuth tokens are only accepted by the /mcp-server/http endpoint,
+ * not by the public REST API (/api/v1/ which requires X-N8N-API-KEY).
+ * We validate by sending a lightweight JSON-RPC ping to n8n's MCP endpoint.
+ */
+export class N8nTokenVerifier implements OAuthTokenVerifier {
+  private n8nBaseUrl: string;
+
+  constructor(n8nBaseUrl: string) {
+    // Strip trailing slash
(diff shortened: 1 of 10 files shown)


czlonkowski:
Thanks @rneuplanche — this is a substantial, well-tested piece of work, and making n8n-mcp an OAuth 2.1 resource server (with n8n as the authorization server) is an interesting direction. A few notes since it's still a draft:

**This is a product/security decision** for the maintainer first: whether n8n-mcp should become an OAuth resource server at all. It's not currently — `main` uses inline static-token auth — so this isn't a mechanical merge.

**If it moves forward, it needs rework against current `main`:**
- The branch is ~159 commits behind and conflicts in both shared files (`src/http-server-single-session.ts`, `src/utils/auth.ts`), which were heavily refactored.
- Reconcile with merged **#767**, which already added RFC 6750 `WWW-Authenticate` challenges on 401 — this PR rewrites the same 401 path and would regress/duplicate it.
- `InMemoryOAuthProvider` / `oauth-clients-store.ts` appear to be referenced only by tests, not by the production wiring (which uses `N8nTokenVerifier`) — either wire them in or drop them.
- The `createInterceptRes` Express-response-mocking hack for "both" mode would be cleaner calling the verifier directly.
- Reassess `N8nTokenVerifier` reusing the OAuth bearer token as the n8n REST API key — the public API expects `X-N8N-API-KEY`, and the description notes management tools still need a separate `x-n8n-key`.

Leaving it as a draft for now. Happy to dig in further once there's a decision on whether to take n8n-mcp in this direction.

PR #661, opened 2026-03-24
feat: add MiniMax as LLM provider for community docs generation (default M3)

octo-patch:
## Summary

- Add a **provider preset system** for the documentation generator with built-in support for MiniMax, OpenAI, and Anthropic
- Users can configure cloud LLM providers via a single `N8N_MCP_LLM_PROVIDER` env var (e.g., `minimax`, `openai`)
- Default model for `minimax` preset is **MiniMax-M3** (512K context, 128K max output, image input support); legacy `MiniMax-M2.7` and `MiniMax-M2.7-highspeed` remain accessible via `N8N_MCP_LLM_MODEL` override
- Add **think-tag stripping** (`<think>...</think>`) for reasoning models like MiniMax M2.7 and Qwen
- Add **temperature clamping** to respect provider-specific ranges (MiniMax: 0-1)
- Auto-detect `MINIMAX_API_KEY` in the API key fallback chain

## Changes

| File | Change |
|------|--------|
| `src/community/documentation-generator.ts` | Provider presets (default model `MiniMax-M3`), think-tag stripping, temperature clamping, factory function update |
| `src/community/index.ts` | Export new types and functions |
| `.env.example` | LLM provider configuration section with MiniMax docs |
| `README.md` | Provider preset table (default `MiniMax-M3`), configuration guide, MiniMax acknowledgment |
| `tests/unit/community/llm-provider-presets.test.ts` | Unit tests for presets, clamping, resolution (asserting `MiniMax-M3` default) |
| `tests/unit/community/documentation-generator.test.ts` | Additional think-tag stripping tests |
| `tests/integration/community/llm-provider-integration.test.ts` | Integration tests for end-to-end MiniMax flow (asserting `MiniMax-M3` default) |

## Test plan

- [x] All new+existing tests pass (`npx vitest run`)
- [x] TypeScript type check passes (`npx tsc --noEmit`)
- [x] Provider preset resolution (case-insensitive)
- [x] Temperature clamping within MiniMax range [0, 1]
- [x] Think-tag stripping for single and multi-block responses
- [x] API key auto-detection fallback chain
- [x] Manual config overrides preset defaults (e.g., `N8N_MCP_LLM_MODEL=MiniMax-M2.7-highspeed`)
- [x] Batch documentation generation with MiniMax

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -187,6 +187,28 @@ ENABLE_MULTI_TENANT=false
 # Set to true to automatically generate metadata when running fetch:templates
 # METADATA_GENERATION_ENABLED=false
 
+# ==========================================
+# LLM PROVIDER CONFIGURATION (Community Docs)
+# ==========================================
+# Configure the LLM used for AI-powered community node documentation generation.
+#
+# Option 1: Use a provider preset (recommended for cloud providers)
+# Supported presets: openai, minimax, anthropic
+# N8N_MCP_LLM_PROVIDER=minimax
+#
+# Option 2: Manual configuration (for local servers or custom endpoints)
+# N8N_MCP_LLM_BASE_URL=http://localhost:1234/v1
+# N8N_MCP_LLM_MODEL=qwen3-4b-thinking-2507
+#
+# API key (auto-detected from MINIMAX_API_KEY or OPENAI_API_KEY if not set)
+# N8N_MCP_LLM_API_KEY=
+#
+# MiniMax API key (get from https://platform.minimaxi.com)
+# MINIMAX_API_KEY=
+#
+# Request timeout in milliseconds (default: 60000)
+# N8N_MCP_LLM_TIMEOUT=60000
+
 # ========================================
 # INTEGRATION TESTING CONFIGURATION
 # ========================================
--- a/README.md
+++ b/README.md
@@ -1259,7 +1259,7 @@ Set `N8N_MCP_LLM_PROVIDER` for quick setup with a cloud provider:
 | Provider | Preset Name | Default Model | API Key Env Var |
 |----------|------------|---------------|-----------------|
 | [OpenAI](https://openai.com) | `openai` | `gpt-4o-mini` | `OPENAI_API_KEY` |
-| [MiniMax](https://www.minimaxi.com) | `minimax` | `MiniMax-M2.7` | `MINIMAX_API_KEY` |
+| [MiniMax](https://www.minimaxi.com) | `minimax` | `MiniMax-M3` | `MINIMAX_API_KEY` |
 | [Anthropic](https://anthropic.com) | `anthropic` | `claude-sonnet-4-20250514` | `N8N_MCP_LLM_API_KEY` |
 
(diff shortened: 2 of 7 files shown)


czlonkowski:
Thanks @octo-patch — a provider-preset system for the community-docs generator is a reasonable idea, and the MiniMax base URL (`https://api.minimax.io/v1`) and default model (`MiniMax-M3`) do match MiniMax's current docs. But the branch is ~146 commits behind and was written against an older `main` that has since been hardened, so it both conflicts with and partially reverts merged work. It needs rework before it can merge:

**Blocking**
1. **Security regression (CodeQL #709).** The factory rewrite reintroduces `baseUrl.includes('openai.com') || baseUrl.includes('anthropic.com') || baseUrl.includes('minimax.io')` substring cloud-detection. Current `main` deliberately replaced exactly this with hostname parsing (`new URL(baseUrl).hostname`, suffix checks) to fix `js/incomplete-url-substring-sanitization`. Merging this reverts that fix — e.g. `http://attacker.com/minimax.io/` would be misclassified as a cloud provider. Please layer the preset logic on top of main's hostname check, and extend the allowlist with `host === 'api.minimax.io' || host.endsWith('.minimax.io')` if needed.
2. **Edits a deleted test file.** The diff modifies `tests/unit/community/documentation-generator.test.ts`, which was intentionally removed from `main` (commit `7e314f8`). That hunk can't apply. Drop it; if `<think>`-stripping coverage is still wanted, add it to an existing test file (note `main` already strips `<think>` in `extractJson()`, so de-dupe against that).
3. **Unsound integration tests.** The MiniMax "end-to-end" tests assert `result.error` is undefined but never stub `global.fetch`, while `chatCompletion()` uses raw `fetch` (not the mocked OpenAI client / injected `client`). They'll either throw in CI or hit the real network. Stub `fetch` via `vi.stubGlobal`, drop the `Object.defineProperty(generator,'client',...)` injection, and assert the outbound request shape (URL, `Authorization`, `body.model`/`temperature`).
4. **`anthropic` preset is questionable.** Anthropic's OpenAI-compat layer at `api.anthropic.com/v1/chat/completions` exists but is documented as test/eval-only, and the code also sends non-standard fields (`chat_template_kwargs`, `max_completion_tokens`). Either drop the preset or gate it clearly as experimental and verify against Anthropic's compat docs. (Note: for first-class Claude support the right path is Bedrock/Vertex, not this layer.)

Please rebase onto `main` and address the above; happy to re-review.

octo-patch:
Thanks for the careful review — all four points are fair and I'll rework rather than push back. Plan: rebase onto current `main`, layer the preset logic on top of the hostname-based cloud detection so the CodeQL `incomplete-url-substring-sanitization` fix stays intact, drop the hunk touching the deleted `documentation-generator.test.ts`, and properly `vi.stubGlobal('fetch', ...)` in the integration tests instead of injecting the client. I'll also drop the `anthropic` preset rather than ship it half-supported. Will re-request review once it's rebased and green.

PR #662, opened 2026-03-24
fix: implement __patch_find_replace in setNestedProperty

### LINKED ISSUE #642: __patch_find_replace in updateNode corrupts jsCode instead of applying patch
When using n8n_update_partial_workflow with updateNode and including __patch_find_replace in the updates object to modify parameters.jsCode, the find/replace is not applied. Instead, the entire jsCode value is replaced with the literal patch object {"__patch_find_replace": [...]}.

At runtime, n8n serializes this as [object Object], causing:

SyntaxError: Unexpected identifier 'Object'

**To reproduce**

1. Call n8n_update_partial_workflow with an operation like:

{
  "type": "updateNode",
  "name": "My Code Node",
  "updates": {
    "parameters.jsCode": {
      "__patch_find_replace": [
        { "find": "old text", "replace": "new text" }
      ]
    }
  }
}
2. Fetch the workflow and inspect the node's parameters.jsCode
3. The jsCode is now the patch object itself, not the patched code

**Expected behavior**
The find/replace should be applied to the existing jsCode string and the result stored as the new jsCode string.
**Workaround**
Fetch the full jsCode first, apply replacements locally, then push the complete string:

{
  "type": "updateNode",
  "name": "My Code Node",
  "updates": {
    "parameters.jsCode": "<full corrected code string>"
  }
}

**Environment**
n8n-mcp version: 2.37.3
n8n version: 2.11.4

JasonOA888:
## Summary

Fixes #642

## Problem

When using `updateNode` with `__patch_find_replace` in updates, the patch object was being set as the literal value instead of applying the find/replace operation to the existing string.

```javascript
// This would corrupt jsCode
n8n_update_partial_workflow({
  operations: [{
    type: "updateNode",
    name: "My Code Node",
    updates: {
      "parameters.jsCode": {
        "__patch_find_replace": [
          { "find": "old text", "replace": "new text" }
        ]
      }
    }
  }]
})
```

Result: `jsCode` becomes `[object Object]` → `SyntaxError`

## Solution

Modified `setNestedProperty()` to detect `__patch_find_replace` objects and apply the patches to existing string values:

```typescript
if (value && typeof value === 'object' && '__patch_find_replace' in value) {
  const patches = value.__patch_find_replace;
  if (Array.isArray(patches) && typeof current[finalKey] === 'string') {
    let result = current[finalKey];
    for (const patch of patches) {
      if (patch.find && typeof patch.replace === 'string') {
        result = result.split(patch.find).join(patch.replace);
      }
    }
    current[finalKey] = result;
    return;
  }
}
```

## Testing

- Existing string values get patched correctly
- Multiple patches are applied in order
- Non-string values fall back to direct assignment

### DIFF
--- a/src/services/workflow-diff-engine.ts
+++ b/src/services/workflow-diff-engine.ts
@@ -1280,6 +1280,22 @@ export class WorkflowDiffEngine {
     }
 
     const finalKey = keys[keys.length - 1];
+    
+    // Handle __patch_find_replace for string values (Issue #642)
+    if (value && typeof value === 'object' && '__patch_find_replace' in value) {
+      const patches = value.__patch_find_replace;
+      if (Array.isArray(patches) && typeof current[finalKey] === 'string') {
+        let result = current[finalKey];
+        for (const patch of patches) {
+          if (patch.find && typeof patch.replace === 'string') {
+            result = result.split(patch.find).join(patch.replace);
+          }
+        }
+        current[finalKey] = result;
+        return;
+      }
+    }
+    
     if (value === null) {
       delete current[finalKey];
     } else {


czlonkowski:
Closing — this was implemented in PR #672 (merged as commit 6be9ffa). We added full `__patch_find_replace` support in `applyUpdateNode()` with validation, sequential patch application, and graceful warnings when find strings aren't matched. See the [changelog for v2.41.1](https://github.com/czlonkowski/n8n-mcp/blob/main/CHANGELOG.md#2411---2026-03-27). Thank you for the contribution\!

PR #667, opened 2026-03-26
fix: normalize If node operator names (isNotEmpty -> notEmpty)

### LINKED ISSUE #665: n8n_update_partial_workflow uses wrong If node operator names — isNotEmpty vs notEmpty
Bug
When using n8n_update_partial_workflow to update an If node (v2.3) condition, the MCP tool's validator and auto-sanitization use operator names that differ from what the n8n execution engine expects. Updates report success and the n8n API shows the correct saved definition, but the execution engine silently ignores the unrecognized operator and defaults to false.

Reproduction
Have a workflow with an If node (typeVersion 2.3) that checks if a field is not empty
Use n8n_update_partial_workflow to set the condition:

{
  "type": "updateNode",
  "nodeName": "My If Node",
  "updates": {
    "parameters.conditions.conditions": [{
      "id": "check",
      "leftValue": "={{ $json.my_field }}",
      "rightValue": "",
      "operator": {
        "type": "number",
        "operation": "isNotEmpty",
        "singleValue": true
      }
    }]
  }
}
MCP reports success: true, saved: true
GET the workflow via n8n API — shows the updated condition
Execution always routes items to the FALSE branch, regardless of the field value
Root cause
The MCP validator defines isNotEmpty as a valid unary operator, but the n8n execution engine expects notEmpty. Conversely, the validator rejects notEmpty as a "binary operator" that should not have singleValue: true — but notEmpty is what the n8n UI generates for the same "is not empty" check.

MCP validator says:

Unary (requires singleValue: true): isEmpty, isNotEmpty, true, false, isNumeric
Binary (must NOT have singleValue: true): notEmpty, ...
n8n execution engine expects:

notEmpty with singleValue: true (this is what the n8n UI produces)
isNotEmpty is not recognized — silently treated as false
Impact
Silent failure: No error at save time, no error at execution time. The workflow runs "successfully" but the condition always evaluates false.
Hard to diagnose: The saved workflow definition looks correct. Only inspecting execution output data reveals the wrong branch is taken.
Using notEmpty (the correct name) via MCP is blocked by the validator, which rejects it as a binary operator. The only workaround is bypassing MCP and using the n8n REST API directly.
Affected operators (likely)
If the naming convention mismatch is systematic, these pairs may also be affected:

MCP validator name	n8n engine name
isNotEmpty	notEmpty
isEmpty	empty
Environment
n8n-mcp: v2.37.3
n8n Cloud: v1.x (cloud-hosted)
If node typeVersion: 2.3

bluzername:
## Problem

When using `n8n_update_partial_workflow` to update an If node condition with `isNotEmpty` operator, the workflow saves successfully but always evaluates to FALSE at runtime. No error anywhere - completely silent failure.

The root cause is that the MCP validator uses `isNotEmpty`/`isEmpty` as operator names, but the n8n execution engine expects `notEmpty`/`empty`. The n8n UI also generates `notEmpty`/`empty`. Since the engine doesnt recognize `isNotEmpty`, it silently defaults to false.

I spent a lot of time debugging this because the saved workflow looks perfectly correct in the API response.

## What I changed

In `src/services/node-sanitizer.ts`:

1. **Added `OPERATOR_ALIASES` map** that translates LLM-friendly names to what n8n engine expects:
   - `isNotEmpty` -> `notEmpty`
   - `isEmpty` -> `empty`

2. **Added `normalizeOperationName()` function** that applies the alias mapping

3. **Applied normalization in `sanitizeOperator()`** - after the operator structure is fixed but before singleValue is set. This way the operation name is correct by the time the unary check runs.

4. **Updated `inferDataType()`** - removed `isEmpty`/`isNotEmpty` from the boolean operations list since after normalization they become `empty`/`notEmpty` which are object-type operations. The comment explains why.

## How it works now

```
LLM generates: {type: "number", operation: "isNotEmpty", singleValue: true}
Sanitizer normalizes to: {type: "number", operation: "notEmpty", singleValue: true}
n8n engine recognizes "notEmpty" and evaluates correctly
```

Before this fix, `isNotEmpty` was passed through unchanged and the engine silently ignored it.

Closes #665

### DIFF
--- a/src/services/node-sanitizer.ts
+++ b/src/services/node-sanitizer.ts
@@ -177,6 +177,15 @@ function sanitizeOperator(operator: any): any {
     }
   }
 
+  // Normalize operator names to what n8n engine expects (e.g. isNotEmpty -> notEmpty)
+  if (sanitized.operation) {
+    const normalized = normalizeOperationName(sanitized.operation);
+    if (normalized !== sanitized.operation) {
+      logger.debug(`Normalizing operator: "${sanitized.operation}" -> "${normalized}"`);
+      sanitized.operation = normalized;
+    }
+  }
+
   // Set singleValue based on operator type
   if (sanitized.operation) {
     if (isUnaryOperator(sanitized.operation)) {
@@ -190,6 +199,20 @@ function sanitizeOperator(operator: any): any {
   }
 
   return sanitized;
+}
+
+/**
+ * Normalize operator names to what the n8n execution engine expects.
+ * The n8n UI generates "notEmpty"/"empty" but LLMs often produce
+ * "isNotEmpty"/"isEmpty" which the engine silently ignores (#665).
+ */
+const OPERATOR_ALIASES: Record<string, string> = {
+  isNotEmpty: 'notEmpty',
+  isEmpty: 'empty',
+};
+
+function normalizeOperationName(operation: string): string {
+  return OPERATOR_ALIASES[operation] ?? operation;
 }
 
 /**
@@ -207,7 +230,8 @@ function isOperationName(value: string): boolean {
  */
 function inferDataType(operation: string): string {
   // Boolean operations
-  const booleanOps = ['true', 'false', 'isEmpty', 'isNotEmpty'];
+  // Note: isEmpty/isNotEmpty are normalized to empty/notEmpty (object-type) by sanitizer
+  const booleanOps = ['true', 'false'];
   if (booleanOps.includes(operation)) {
     return 'boolean';
   }


czlonkowski:
Closing — this was fixed in PR #672 (merged as commit 6be9ffa). We replaced `isNotEmpty`/`isEmpty` with `notEmpty`/`empty` across all validators, sanitizer, docs, and error messages, plus added auto-correction in the sanitizer. See the [changelog for v2.41.1](https://github.com/czlonkowski/n8n-mcp/blob/main/CHANGELOG.md#2411---2026-03-27). Thank you for the contribution!

Issue #617, opened 2026-03-03
SSE reconnection loop: POST /mcp handler doesn't route to SSEServerTransport


KengeterSmartHome:
## Bug Description

SSE connections enter a rapid reconnection loop because `POST /mcp` messages from SSE clients are never routed to the `SSEServerTransport.handlePostMessage()` method. This causes the client to never receive a response on its SSE stream, leading to immediate reconnection attempts that quickly exhaust the rate limiter.

## Steps to Reproduce

1. Connect an MCP client via SSE (`GET /mcp` with `Accept: text/event-stream`)
2. Client sends `initialize` via `POST /mcp`
3. Observe rapid session cycling in logs

## Observed Behavior

```
[INFO] SSE stream request received - establishing SSE connection
[INFO] Created new SSE session successfully { sessionId: '2aed...' }
[INFO] SSE connection established successfully
[WARN] Rate limit exceeded { ip: '...', event: 'rate_limit' }
[INFO] Closing previous session for SSE { sessionId: '2aed...' }
[INFO] Created new SSE session successfully { sessionId: 'ef2d...' }
[INFO] Closing previous session for SSE { sessionId: 'ef2d...' }
[INFO] Created new SSE session successfully { sessionId: '5a76...' }
// ... loop continues, 4+ sessions created within 2 seconds
```

## Root Cause Analysis

The issue is in `src/http-server-single-session.ts`:

### 1. Missing SSE POST routing (critical)

`resetSessionSSE()` (line ~706) creates an `SSEServerTransport` with `'/mcp'` as the message endpoint, storing it in `this.session`. However, the `POST /mcp` handler only processes `StreamableHTTPServerTransport` sessions via `this.handleRequest()`, which exclusively looks in `this.transports` (the StreamableHTTP sessions map).

The SSE session stored in `this.session` is **invisible** to the POST handler. `SSEServerTransport.handlePostMessage()` is never called.

**Chain of events:**
1. Client connects via SSE → `GET /mcp` → `resetSessionSSE()` creates session in `this.session`
2. Client sends `initialize` → `POST /mcp` → `handleRequest()` only checks `this.transports` (empty for SSE)
3. Since no `mcp-session-id` header exists and it's an initialize request, a new `StreamableHTTPServerTransport` is created instead
4. Client receives no response on the SSE stream → reconnects → closes previous session → loop

### 2. Rate limiter compounds the issue

[...]

PR #691, opened 2026-04-01
Support legacy auth env config in HTTP server

### LINKED ISSUE #509: AUTH_MAX_ATTEMPTS and AUTH_ENABLED=false environment variables are ignored
Hi!

The `n8n-mcp` server ignores the `AUTH_MAX_ATTEMPTS` and `AUTH_ENABLED` environment variables, causing persistent HTTP 429 errors with the message "Too many authentication attempts" even when these variables are explicitly configured.

## Environment

- **n8n-mcp image:** `ghcr.io/czlonkowski/n8n-mcp:latest`
- **Platform:** Docker on Linux (Debian)
- **Client:** `mcp-remote 0.1.36` via Cursor IDE
- **Transport:** HTTP-only (`--transport http-only`)

## Configuration

```yaml
n8n-mcp:
  image: ghcr.io/czlonkowski/n8n-mcp:latest
  environment:
    - PORT=3001
    - MCP_MODE=http
    - AUTH_TOKEN=<token>
    - AUTH_MAX_ATTEMPTS=999999999
    - AUTH_ENABLED=false
    - RATE_LIMIT_ENABLED=false
    - MAX_SESSIONS=1000
```

## Verified

Environment variables are correctly set inside the container:

```bash
$ docker exec n8n-n8n-mcp-1 env | grep -E "(MAX|RATE|AUTH)"
AUTH_MAX_ATTEMPTS=999999999
AUTH_ENABLED=false
RATE_LIMIT_ENABLED=false
AUTH_TOKEN=<token>
```

However, server logs still show:
```
Session Limits: 100 max sessions, 30min timeout
```

## Error

```json
{"jsonrpc":"2.0","error":{"code":-32000,"message":"Too many authentication attempts"},"id":null}
```

This occurs after just 2-3 connection attempts, despite `AUTH_MAX_ATTEMPTS=999999999`.

## Expected Behavior

1. `AUTH_ENABLED=false` should disable authentication entirely
2. `AUTH_MAX_ATTEMPTS=999999999` should allow virtually unlimited attempts
3. `MAX_SESSIONS=1000` should be respected instead of hardcoded 100

## Additional Notes

- Removing `AUTH_TOKEN` causes the container to crash (exit code 1)
- Setting `AUTH_TOKEN=""` also crashes the container
- The rate limit resets after ~2 minutes, allowing temporary reconnection

Tejasvp24:
This fixes #509.

Summary:
- honor `AUTH_ENABLED=false` in the single-session HTTP server and skip bearer auth when explicitly disabled
- honor `RATE_LIMIT_ENABLED=false` for auth rate limiting
- accept legacy `MAX_SESSIONS` and `AUTH_MAX_ATTEMPTS` aliases alongside the current env names
- document the compatibility behavior in `.env.example`
- add unit coverage for the env compatibility paths

Verification:
- `npx tsc --noEmit --pretty false`
- `npx vitest run tests/unit/http-server-n8n-mode.test.ts --coverage.enabled=false`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -77,10 +77,16 @@ AUTH_TOKEN=your-secure-token-here
 
 # Rate Limiting Configuration
 # Protects authentication endpoint from brute force attacks
+# Set RATE_LIMIT_ENABLED=false to disable auth rate limiting entirely
 # Window: Time period in milliseconds (default: 900000 = 15 minutes)
 # Max: Maximum authentication attempts per IP within window (default: 20)
+# Legacy alias supported: AUTH_MAX_ATTEMPTS
 # AUTH_RATE_LIMIT_WINDOW=900000
 # AUTH_RATE_LIMIT_MAX=20
+
+# Authentication Configuration
+# Set AUTH_ENABLED=false to disable bearer-token auth for HTTP mode
+# AUTH_ENABLED=true
 
 # SSRF Protection Mode
 # Prevents webhooks from accessing internal networks and cloud metadata
--- a/src/http-server-single-session.ts
+++ b/src/http-server-single-session.ts
@@ -42,8 +42,33 @@ interface MultiTenantHeaders {
   'x-session-id'?: string;
 }
 
-// Session management constants
-const MAX_SESSIONS = Math.max(1, parseInt(process.env.N8N_MCP_MAX_SESSIONS || '100', 10));
+function parseBooleanEnv(...names: string[]): boolean | undefined {
+  for (const name of names) {
+    const value = process.env[name];
+    if (value === undefined) continue;
+
+    const normalized = value.trim().toLowerCase();
+    if (['true', '1', 'yes', 'on'].includes(normalized)) return true;
+    if (['false', '0', 'no', 'off'].includes(normalized)) return false;
+  }
+
+  return undefined;
+}
+
+function parseNumberEnv(defaultValue: number, minimum: number, ...names: string[]): number {
+  for (const name of names) {
+    const value = process.env[name];
+    if (value === undefined) continue;
+
+    const parsed = parseInt(value, 10);
+    if (!Number.isNaN(parsed)) {
+      return Math.max(minimum, parsed);
+    }
+  }
+
+  return defaultValue;
+}
+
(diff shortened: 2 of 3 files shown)


czlonkowski:
Thanks @Tejasvp24 — honoring `AUTH_ENABLED`, `RATE_LIMIT_ENABLED`, and the legacy `MAX_SESSIONS` / `AUTH_MAX_ATTEMPTS` aliases (issue #509) is worth doing, and the `parseBooleanEnv`/`parseNumberEnv` helpers and instance-field refactor are clean in isolation. But this is a security-sensitive change and it was written against a stale base, so it needs rework before merge:

**Blocking**
1. **Auth bypass is wired to the wrong place (security).** The PR gates auth on `this.authEnabled` inside an inline auth block in `POST /mcp`. Current `main` no longer has that inline block — it calls a shared `authenticateRequest()` from **five** endpoints (POST/GET/DELETE `/mcp`, GET `/sse`, POST `/messages`). After rebase your edit either won't apply or, worse, would disable auth only for `POST /mcp` while the other four keep enforcing it — an inconsistent, partially-open server. Please move the bypass into `authenticateRequest()` itself: `if (!this.authEnabled) return true;` (with a log), so it applies uniformly.
2. **No `NODE_ENV` guard on disabling auth.** `AUTH_ENABLED=false` unconditionally nulls the token with no production stop. The repo already throws in production for a default token; disabling auth entirely is strictly more dangerous. Please `throw` when `NODE_ENV==='production'` (and `logger.warn` otherwise).
3. **Rate-limiter regression (#617).** The rewritten `authLimiter` drops `skipSuccessfulRequests: true` (added in #617) and re-types the old options. Gate at the single shared declaration instead — `const authLimiter = this.rateLimitEnabled ? rateLimit({...keep all main options...}) : (req,res,next)=>next();` — keyed only on `rateLimitEnabled` (decouple it from `authEnabled`), kept before all route registrations so all five endpoints honor it.

**Tests**
Add: `RATE_LIMIT_ENABLED=false`; that `AUTH_ENABLED=false` bypasses **all five** endpoints (not just POST /mcp); that it's refused when `NODE_ENV=production`; and a regression test that auth is still enforced by default.

Please rebase onto current `main` and re-implement against the `authenticateRequest()` / shared-limiter architecture. Happy to re-review.

Issue #628, opened 2026-03-11
MCP initialization produces multiple "Unexpected token is not valid JSON" errors on every new chat


SolarFab:
Description:
On every new chat session, Claude displays a flood of yellow warning notifications:

MCP n8n-local: Unexpected token 'r', " refCount: 1" is not valid JSON
MCP n8n-local: Unexpected token 'c', " clientVersion"... is not valid JSON
MCP n8n-local: Unexpected token 'v', " version: '"... is not valid JSON
MCP n8n-local: Unexpected token 'i', " isN8nClient"... is not valid JSON
MCP n8n-local: Unexpected token 'r', " reasoning:"... is not valid JSON
MCP n8n-local: Unexpected token 'r', " response: {"... is not valid JSON
MCP n8n-local: Unexpected token 'p', " protocolVersion"... is not valid JSON
MCP n8n-local: Unexpected token 'c', " capabilities"... is not valid JSON
MCP n8n-local: Unexpected token 's', " serverInfo"... is not valid JSON

Environment:

n8n-mcp version: 2.36.1 (latest)
Platform: macOS, Claude Desktop app
n8n version: 2.10.4 running on a remote Linux server
Connection: via Tailscale (HTTP, not HTTPS)
n8n connection itself works fine (health check returns success)

Expected behavior: MCP server initializes silently without errors.
Actual behavior: During the MCP handshake/initialization, the server sends non-JSON formatted metadata fields which Claude's interface cannot parse, resulting in multiple error notifications on every new chat.
Note: These errors are purely cosmetic — the connection works correctly after initialization.

Issue #555, opened 2026-01-24
Windows: MCP stdio connection times out due to log output on stdout


martinfisch3r:
Description
When using n8n-mcp with Claude Code VSCode extension on Windows, the MCP connection times out after 30 seconds. The issue is caused by log messages being written to stdout, which corrupts the JSON-RPC protocol used by MCP stdio transport.

Environment
OS: Windows 11
Node.js: v24.12.0
n8n-mcp: latest (via npx)
MCP Client: Claude Code VSCode Extension v2.1.19

Configuration Used
{
  "mcpServers": {
    "n8n-mcp": {
      "command": "npx",
      "args": ["n8n-mcp"],
      "env": {
        "MCP_MODE": "stdio",
        "LOG_LEVEL": "error",
        "DISABLE_CONSOLE_OUTPUT": "true",
        "N8N_API_URL": "https://example.app.n8n.cloud/",
        "N8N_API_KEY": "xxx"
      }
    }
  }
}

Steps to Reproduce
Configure n8n-mcp in .mcp.json with the settings above
Open Claude Code in VSCode on Windows
Observe the Output logs
Expected Behavior
MCP server should connect successfully with no log output on stdout when DISABLE_CONSOLE_OUTPUT=true and LOG_LEVEL=error are set.

Actual Behavior
The connection times out after 30 seconds:
MCP server "n8n-mcp": Starting connection with timeout of 30000ms
MCP server "n8n-mcp": Connection timeout triggered after 30040ms
MCP server "n8n-mcp": Connection failed: connection timed out after 30000ms

When running npx n8n-mcp manually, log output is visible despite the environment variables:
[n8n-mcp] [INFO] Initializing shared database connection...
[n8n-mcp] [INFO] Node.js version: v24.12.0
[n8n-mcp] [INFO] Platform: win32 x64
[n8n-mcp] [WARN] Failed to initialize better-sqlite3, falling back to sql.js...
[n8n-mcp] [INFO] MCP server initialized with 7 tools
[n8n-mcp] [INFO] n8n Documentation MCP Server running on stdio transport
These log messages appear to go to stdout instead of stderr, corrupting the MCP JSON-RPC communication.

Attempted Workarounds (Failed)
CMD wrapper script with 2>NUL redirection
PowerShell wrapper with 2>$null redirection
Node.js wrapper suppressing console methods

Working Workaround
Using the hosted service at dashboard.n8n-mcp.com with HTTP transport works correctly.

[...]

PR #694, opened 2026-04-01
fix: use stdio-wrapper.js as default bin entry point

### LINKED ISSUE #693: Default bin entry (index.js) corrupts stdio transport — should use stdio-wrapper.js
## Bug

The default bin entry point (`dist/mcp/index.js`) writes INFO-level logs to **stdout** via `console.log()`, corrupting the JSON-RPC MCP transport. This causes MCP clients (Claude Code, Claude Desktop, VS Code extension) to fail tool registration despite reporting "Connected."

## Root Cause

The logger (`dist/utils/logger.js`) uses `console.log()` for INFO level. Only `console.error()` writes to stderr; the others pollute stdout.

The package already contains the correct fix at `dist/mcp/stdio-wrapper.js` — it silences all console methods, intercepts `process.stdout.write` to only pass JSON-RPC messages, and redirects everything else to stderr. However, `package.json` wires the bin entry to `dist/mcp/index.js` instead of `dist/mcp/stdio-wrapper.js`.

## Prior Reports

- #555 (Windows stdio timeout due to log output on stdout) — closed, no fix merged
- #628 (JSON parse errors from log leakage on every new chat) — closed, no fix merged
- Bug persists in v2.44.1

## Reproduction

```bash
echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"test","version":"1.0"}}}' | \
  N8N_API_URL=https://example.com N8N_API_KEY=test npx n8n-mcp@2.44.1 2>/dev/null
```

Expected: only `{"jsonrpc":...}` lines on stdout.
Actual: `[n8n-mcp] [INFO] ...` lines mixed in, breaking JSON-RPC parsing.

## Proposed Fix

One-line change in `package.json`:

```diff
-  "bin": { "n8n-mcp": "./dist/mcp/index.js" }
+  "bin": { "n8n-mcp": "./dist/mcp/stdio-wrapper.js" }
```

The `stdio-wrapper.js` already exists and does the right thing — it just needs to be the default entry point.

## Workaround

Set `LOG_LEVEL=error` and `DISABLE_CONSOLE_OUTPUT=true` as environment variables. This suppresses the logs but is fragile — any new log statement at INFO level re-breaks the transport.

gjenkins20:
## Summary

Changes the `bin` entry in `package.json` from `./dist/mcp/index.js` to `./dist/mcp/stdio-wrapper.js`.

The current default entry point (`index.js`) writes INFO-level logs to **stdout** via `console.log()`, which corrupts the JSON-RPC MCP transport. The `stdio-wrapper.js` already exists in the package and properly:
- Silences all `console.*` methods
- Intercepts `process.stdout.write` to only pass JSON-RPC messages
- Redirects non-JSONRPC output to stderr

This is a one-line change — no new code, just wiring the correct entry point.

Fixes #693. Related: #555, #628.

## Test plan

- [ ] Run `npx n8n-mcp` with default env (no `LOG_LEVEL` or `DISABLE_CONSOLE_OUTPUT`) and verify no `[n8n-mcp] [INFO]` lines appear on stdout
- [ ] Verify MCP stdio transport initializes cleanly with Claude Desktop / Claude Code
- [ ] Verify HTTP mode still works (unaffected — uses different entry path)

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     }
   },
   "bin": {
-    "n8n-mcp": "./dist/mcp/index.js"
+    "n8n-mcp": "./dist/mcp/stdio-wrapper.js"
   },
   "scripts": {
     "build": "tsc -p tsconfig.build.json",


czlonkowski:
Hey @gjenkins20, thanks for this PR! You were spot on — the `stdio-wrapper.js` is the correct entry point for MCP stdio mode.

We merged the fix in #695 (v2.45.1) which included your change along with a credential preservation fix. Your contribution is credited in the changelog and commit.

Closing this as the fix is already on main. Thanks for catching this!

PR #710, opened 2026-04-08
docs: Fix macOS path formatting in README.md

fureigh:
Adds a missing escape character to correct the configuration file location.

Thanks for all of your work!

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ Add to Claude Desktop config:
 > **Note**: npx will download and run the latest version automatically. The package includes a pre-built database with all n8n node information.
 
 **Configuration file locations:**
-- **macOS**: `~/Library/Application Support/Claude/claude_desktop_config.json`
+- **macOS**: `~/Library/Application\ Support/Claude/claude_desktop_config.json`
 - **Windows**: `%APPDATA%\Claude\claude_desktop_config.json`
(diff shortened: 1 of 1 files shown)


czlonkowski (changes requested):
Thanks for spotting an inconsistency! However, this section lists *filesystem paths* (note Windows uses `%APPDATA%` without escaping), not shell commands — users typically open this file via Finder, a text editor, or `code ~/Library/...` where the unescaped path is correct. Escaping the space would break "Go to Folder" in Finder and similar GUI flows.

Could we either close this, or would you like to instead add a note like *"if pasting into a terminal, wrap the path in quotes"*?

czlonkowski:
Hi @fureigh, thank you for taking the time to spot this and open a PR - attention to detail on the docs is genuinely appreciated.

As discussed in the review, this section lists *filesystem paths* rather than shell commands (note the Windows entry uses an unescaped `%APPDATA%` for the same reason). Users typically open this file through Finder, a text editor, or `code ~/Library/...`, where the unescaped space is correct; backslash-escaping it would actually break Finder's "Go to Folder" and similar GUI flows.

On top of that, the README has since been restructured and the "Configuration file locations" section these lines lived in has been removed entirely, so the change no longer applies to the current file.

For both of those reasons I'm going to close this one out - but thank you again for contributing and for caring about the docs. Please don't let this discourage you from opening future PRs!

PR #719, opened 2026-04-09
feat(mcp): add per-operation tool filtering via DISABLED_TOOL_OPERATIONS

### LINKED ISSUE #714: Per-operation tool filtering for read-only deployments (DISABLED_TOOL_OPERATIONS)
## Background

Several users running n8n-mcp in enterprise / governance-sensitive contexts want a true read-only mode but are blocked by tools that bundle read and write operations under a single tool name. The existing `DISABLED_TOOLS` env var only operates at the tool level, so disabling a destructive branch means losing the read branch as well.

Discussed in #564 (by @mahmoudnaif, with a strong governance writeup from @christiansousadev).

## The problem

Two tools mix read and destructive operations under one name:

- **`n8n_workflow_versions`**
  - Read: `list`, `get`
  - Write/destructive: `rollback`, `delete`, `prune`, `truncate`

- **`n8n_executions`**
  - Read: `list`, `get`
  - Destructive: `delete`

Today the only way to block the destructive branches via n8n-mcp is to disable the entire tool, which also kills the valuable read paths. Tool annotations (`destructiveHint: true`) exist but are metadata only — nothing in the dispatch layer enforces them.

The robust workaround today is n8n-side API key scoping (RBAC), which is the right pattern for production but does not give clients a way to express the same intent at the MCP layer.

## Proposed feature

Add an env var that filters operations *within* a tool at dispatch time, before the handler runs:

```
DISABLED_TOOL_OPERATIONS=n8n_workflow_versions:delete,rollback,prune,truncate;n8n_executions:delete
```

Format sketch (open to bikeshedding):
- Semicolon-separated list of `<tool_name>:<comma_separated_operations>`
- Operation names match the existing `action` / `mode` enum values in each tools schema
- Disabled operations should also be advertised as such in `tools_documentation` output so the LLM does not waste a turn trying them

## Acceptance criteria

- [ ] New env var parsed at startup with the same safety limits as `DISABLED_TOOLS` (size cap, count cap, validation against known tool/operation names)
- [ ] Dispatch layer in `src/mcp/server.ts` checks the disabled-operation set before calling the handler and returns a clear error (something like `Operation 'delete' on tool 'n8n_executions' is disabled by server policy`)
- [ ] `tools_documentation` reflects the disabled operations so they are visible to clients
- [ ] Unit tests covering: parser edge cases, dispatch enforcement for both `n8n_workflow_versions` and `n8n_executions`, interaction with existing `DISABLED_TOOLS`
- [ ] README / `docs/` section explaining the read-only deployment recipe (this env var + n8n API key RBAC)

## Out of scope

- Full RBAC inside n8n-mcp — that lives in the n8n API
- Per-user / per-session policy — this is a server-level config knob

## Notes

- The implementation is straightforward: a Set lookup in the dispatch path. Cheap at runtime.
- Happy to review a PR. Tagging @christiansousadev and @mahmoudnaif from the discussion in case either of you wants to take a swing at it.

Concieved by Romuald Członkowski — https://www.aiadvisors.pl/en

mahmoudnaif:
# PR: Per-operation tool filtering via `DISABLED_TOOL_OPERATIONS` — Closes #714

## What this does

Introduces a new environment variable `DISABLED_TOOL_OPERATIONS` that lets operators disable specific operations within a tool at dispatch time, without losing the tool's read paths. This unblocks the read-only deployment use case described in #714.

```
DISABLED_TOOL_OPERATIONS=n8n_workflow_versions:delete,rollback,prune,truncate;n8n_executions:delete
```

Format: semicolon-separated list of `<tool_name>:<comma_separated_operations>`. Operation names match the existing `action` / `mode` enum values in each tool's schema.

## Why

`DISABLED_TOOLS` operates at tool level — disabling `n8n_executions` kills `list` and `get` along with `delete`. Governance-sensitive deployments need a finer knob. This PR adds that knob with minimal surface area and zero runtime overhead after the first request.

### `src/mcp/server.ts`
- **`getDisabledToolOperations()`** — parses `DISABLED_TOOL_OPERATIONS` at first use, caches the result. Same safety limits as `getDisabledTools()`: 10KB env var cap, 50-entry cap.
- **`buildFilteredToolDefinitions()`** — called once on first parse. Deep-clones only the affected tool definitions and filters the operation `enum` array so the LLM never sees disabled ops in `ListTools`.
- **`ListToolsRequestSchema` handler** — applies pre-built filtered definitions via a Map lookup. Zero cloning cost per request.
- **`CallToolRequestSchema` handler** — checks disabled operations before calling the handler. Returns a structured `OPERATION_DISABLED` error.
  > **Note (behaviour change):** also fixes the existing `DISABLED_TOOLS` response to include `isError: true` — previously the response was a content-only envelope; this aligns it with the correct MCP protocol shape and with the new `OPERATION_DISABLED` response. Integrations checking `response.isError` directly will now see `true` instead of `undefined`.
- **`executeTool()`** — secondary defense-in-depth guard, mirrors the existing `DISABLED_TOOLS` guard pattern.

### `src/mcp/tools-n8n-manager.ts`
- **`TOOL_OPERATION_PARAM`** — exported map of `toolName -> argument key`. Necessary because eligible tools use different argument names for their operation selector: `n8n_executions` uses `action`, `n8n_workflow_versions` uses `mode`. The dispatch guard needs to know which argument to inspect per tool. Co-located with the tool definitions so adding a future eligible tool is a single-file change.

### `src/mcp/tools-documentation.ts`
- **`getToolDocumentation()`** — accepts an optional `disabledOperations?: Set<string>` parameter. When set, prepends a clear server policy notice to both essentials and full output.
- **`getToolsOverview()`** — accepts an optional `disabledToolOps?: Map<string, Set<string>>` parameter. When set, appends a `## Server Policy: Disabled Operations` section.

### `tests/unit/mcp/disabled-tool-operations.test.ts`
- 30 tests across 6 groups: parser edge cases, dispatch enforcement for `n8n_executions`, dispatch enforcement for `n8n_workflow_versions`, interaction with `DISABLED_TOOLS`, schema filtering/mutation safety, and `tools_documentation` notice.

### Docs and configuration
- **`README.md`** — Read-Only Deployment section added under Available MCP Tools, covering both `DISABLED_TOOLS` (purely write tools) and `DISABLED_TOOL_OPERATIONS` (mixed-operation tools).
- **`docs/HTTP_DEPLOYMENT.md`** — Read-Only Deployment Recipe section added under Security Best Practices, with the full two-layer setup guide.
- **`.env.example`** — `DISABLED_TOOL_OPERATIONS` block added after `DISABLED_TOOLS` with format, eligible tools, and the read-only recipe example.

## Design decisions worth discussing

**Documentation approach**: disabled operations are surfaced as a plain text notice at the top of the `tools_documentation` output, not by filtering out tips/examples. Filtering free-form text strings is fragile — the notice alone is sufficient for the LLM to understand the constraint.

**Unknown operation names**: if a configured operation doesn't exist in the tool's enum (e.g. a typo), both the schema layer and the dispatch layer are silent no-ops — the enum filter has nothing to remove, and the dispatch guard will never match. The description notice will include the configured name as-is. This is intentionally consistent with how `DISABLED_TOOLS` handles unknown tool names — user's responsibility, server stays stable.

**Caching**: `getDisabledToolOperations()` is lazy — it initialises on the first MCP request (either `ListTools` or `CallTool`), not at server startup. This is the same pattern as `getDisabledTools()`. The comment in the code says so explicitly.

**Scope**: only `n8n_executions` and `n8n_workflow_versions` are eligible today. Adding a new tool requires one line in `TOOL_OPERATION_PARAM` in `tools-n8n-manager.ts`.

## Open questions / suggestions welcome

- Should unknown operation names in the env var be silently ignored (current) or emit a `logger.warn` at parse time listing valid alternatives? Hard failure at startup is probably too strict for a config knob.

- `TOOL_OPERATION_PARAM` exists because the two eligible tools use different argument names for their operation selector (`action` vs `mode`). If future tools are added, this map grows. A cleaner long-term fix would be to standardize the operation selector argument name across all multi-operation tools (e.g. always use `operation`), which would eliminate the need for this map entirely and make per-operation filtering generic.

Conceived by Romuald Czlonkowski — https://www.aiadvisors.pl/en

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -132,10 +132,10 @@ AUTH_TOKEN=your-secure-token-here
 #
 # Eligible tools and their operations:
 #   n8n_executions       — action: get, list, delete
-#   n8n_workflow_versions — mode: list, get, rollback, delete, prune, truncate
+#   n8n_workflow_versions — mode: list, get, rollback, delete, prune
 #
 # Read-only deployment recipe (blocks all destructive operations):
-# DISABLED_TOOL_OPERATIONS=n8n_workflow_versions:delete,rollback,prune,truncate;n8n_executions:delete
+# DISABLED_TOOL_OPERATIONS=n8n_workflow_versions:delete,rollback,prune;n8n_executions:delete
 #
 # Combine with n8n API key RBAC for defence in depth.
 # Default: (empty - all operations enabled)
--- a/.env.example
+++ b/.env.example
@@ -122,6 +122,25 @@ AUTH_TOKEN=your-secure-token-here
 # Default: (empty - all tools enabled)
 # DISABLED_TOOLS=
 
+# Per-Operation Tool Filtering
+# Disable specific operations within a tool without losing its read paths.
+# Useful for sensitive deployments that need read-only access
+# to tools that bundle read and write operations under one name.
+#
+# Format: Semicolon-separated list of <tool_name>:<comma_separated_operations>
+# Operation names match the action / mode enum values in each tool's schema.
+#
+# Eligible tools and their operations:
+#   n8n_executions       — action: get, list, delete
+#   n8n_workflow_versions — mode: list, get, rollback, delete, prune, truncate
+#
+# Read-only deployment recipe (blocks all destructive operations):
+# DISABLED_TOOL_OPERATIONS=n8n_workflow_versions:delete,rollback,prune,truncate;n8n_executions:delete
+#
+# Combine with n8n API key RBAC for defence in depth.
+# Default: (empty - all operations enabled)
+# DISABLED_TOOL_OPERATIONS=
+
 # =========================
 # MULTI-TENANT CONFIGURATION
 # =========================
--- a/README.md
+++ b/README.md
@@ -389,6 +389,22 @@ These tools require `N8N_API_URL` and `N8N_API_KEY` in your configuration.
 #### System Tools
 - **`n8n_health_check`** - Check n8n API connectivity and features
 
+### Read-Only Deployment
+
+For governance-sensitive environments, use both env vars together. Fully disable purely write/destructive tools:
+
+```bash
+DISABLED_TOOLS=n8n_create_workflow,n8n_update_full_workflow,n8n_update_partial_workflow,n8n_delete_workflow,n8n_autofix_workflow,n8n_deploy_template,n8n_test_workflow,n8n_generate_workflow,n8n_manage_credentials,n8n_manage_datatable
+```
+
+For tools that bundle read and write operations under one name, block only the destructive operations while keeping `list` and `get`:
+
+```bash
+DISABLED_TOOL_OPERATIONS=n8n_workflow_versions:delete,rollback,prune;n8n_executions:delete
+```
+
+Combine with a read-only n8n API key (Settings → API in your n8n instance) for defence in depth. See [Read-Only Deployment Recipe](./docs/HTTP_DEPLOYMENT.md#read-only-deployment-recipe) for the full setup guide.
+
 ## Documentation
 
 - [Self-Hosting Guide](./docs/SELF_HOSTING.md) - npx, Docker, Railway, and local installation
--- a/docs/HTTP_DEPLOYMENT.md
+++ b/docs/HTTP_DEPLOYMENT.md
@@ -877,6 +877,32 @@ curl -X POST https://your-server.com/mcp \
 - 🚫 Never expose n8n API directly to the internet
 - ✅ Use MCP server as a security layer
 
+### Read-Only Deployment Recipe
+
+For governance-sensitive environments where the AI agent should be able to read workflow and execution data but must not modify or delete anything, combine two layers of control:
+
+**Layer 1 — MCP layer:**
+
+Some tools are purely write/destructive and should be removed entirely via `DISABLED_TOOLS`:
+
+```bash
+DISABLED_TOOLS=n8n_create_workflow,n8n_update_full_workflow,n8n_update_partial_workflow,n8n_delete_workflow,n8n_autofix_workflow,n8n_deploy_template,n8n_test_workflow,n8n_generate_workflow,n8n_manage_credentials,n8n_manage_datatable
+```
+
+Two tools bundle read and write operations under a single name. Use `DISABLED_TOOL_OPERATIONS` to block only their destructive branches while keeping `list` and `get`:
+
+```bash
+DISABLED_TOOL_OPERATIONS=n8n_workflow_versions:delete,rollback,prune,truncate;n8n_executions:delete
+```
+
+The tool schema presented to the LLM is updated to exclude disabled operations, so the model never attempts them.
+
+**Layer 2 — n8n API key RBAC:**
+
+Scope the `N8N_API_KEY` to read-only permissions in your n8n instance (Settings → API → create a key with read-only scope). This ensures that even if the MCP layer is misconfigured, the n8n API itself will reject destructive requests.
+
+Both layers together provide defence in depth. The MCP layer is a convenience knob; the n8n API RBAC is the authoritative enforcement boundary.
+
 ## 📦 Updates & Maintenance
 
 ### Version Updates
--- a/docs/HTTP_DEPLOYMENT.md
+++ b/docs/HTTP_DEPLOYMENT.md
@@ -892,7 +892,7 @@ DISABLED_TOOLS=n8n_create_workflow,n8n_update_full_workflow,n8n_update_partial_w
 Two tools bundle read and write operations under a single name. Use `DISABLED_TOOL_OPERATIONS` to block only their destructive branches while keeping `list` and `get`:
 
 ```bash
-DISABLED_TOOL_OPERATIONS=n8n_workflow_versions:delete,rollback,prune,truncate;n8n_executions:delete
+DISABLED_TOOL_OPERATIONS=n8n_workflow_versions:delete,rollback,prune;n8n_executions:delete
 ```
 
 The operation parameter enum in the tool schema is updated to exclude disabled values, reducing the likelihood the model attempts them. Any attempt that does reach the server is rejected at dispatch before the handler runs.
--- a/docs/HTTP_DEPLOYMENT.md
+++ b/docs/HTTP_DEPLOYMENT.md
@@ -895,7 +895,7 @@ Two tools bundle read and write operations under a single name. Use `DISABLED_TO
 DISABLED_TOOL_OPERATIONS=n8n_workflow_versions:delete,rollback,prune,truncate;n8n_executions:delete
 ```
 
-The tool schema presented to the LLM is updated to exclude disabled operations, so the model never attempts them.
+The operation parameter enum in the tool schema is updated to exclude disabled values, reducing the likelihood the model attempts them. Any attempt that does reach the server is rejected at dispatch before the handler runs.
 
 **Layer 2 — n8n API key RBAC:**
 
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -15,7 +15,7 @@ import path from 'path';
 import { n8nDocumentationToolsFinal } from './tools';
 import { UIAppRegistry } from './ui';
 import { SkillResourceRegistry } from './skills';
-import { n8nManagementTools, TOOL_OPERATION_PARAM } from './tools-n8n-manager';
+import { n8nManagementTools, TOOL_OPERATION_PARAM, DESTRUCTIVE_TOOL_OPERATIONS } from './tools-n8n-manager';
 import { makeToolsN8nFriendly } from './tools-n8n-friendly';
 import { getWorkflowExampleString } from './workflow-examples';
 import { logger } from '../utils/logger';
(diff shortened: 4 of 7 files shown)


mahmoudnaif:
Hey, wanted to share a thought on `TOOL_OPERATION_PARAM` as a follow-up to the discussion.

The map currently exists because the two eligible tools use different argument names for their operation selector — `action` for `n8n_executions` and `mode` for `n8n_workflow_versions`. The dispatch guard needs to know which argument to inspect per tool, so the map encodes that knowledge explicitly.

An alternative approach would be to let the operator specify the parameter name directly in the env var format, using a third separator character. For example, using `@` to denote the param:

```
DISABLED_TOOL_OPERATIONS=n8n_executions@action:delete;n8n_workflow_versions@mode:delete,rollback,prune,truncate
```

The parser would split each entry on `@` first to extract the tool name and the `param:operations` part, then split on `:` as before. This would make the feature fully generic — any tool with any operation parameter name becomes eligible without touching the codebase. `TOOL_OPERATION_PARAM` would no longer be needed.

This format also naturally extends to tools that need multiple parameters filtered — multiple `@param:operations` segments per tool entry:

```
DISABLED_TOOL_OPERATIONS=n8n_some_tool@action:delete@mode:destructive,dangerous
```

That said, this is where the complexity starts to compound. The parser would need to handle multiple `@` segments per entry, the dispatch guard would need to check multiple params per tool call, and the env var format becomes significantly harder to write correctly by hand. The surface area for misconfiguration grows with each additional separator level.

For the two tools this PR targets, the current map is a net win: the format stays clean and the param names are an implementation detail the operator shouldn't need to know. If the number of eligible tools grows significantly, the dynamic approach becomes more attractive.

Just wanted to put this on the table — happy to hear your thoughts.

diborque2733 (approved):
Nice work @mahmoudnaif. Read through the full diff (`src/mcp/server.ts`, `src/mcp/tools-documentation.ts`, `src/mcp/tools-n8n-manager.ts`, and the 383-line test file) and this is a clean, well-structured implementation of #714. A few observations and a couple of nits that I think can be handled as follow-ups, not blockers.

## Strengths

- **Defense in depth is real, not lip service.** Three independent enforcement points:
  1. `buildFilteredToolDefinitions` strips disabled ops from the `enum` at `ListTools` time so the model never sees them,
  2. the `CallToolRequestSchema` handler rejects with a structured `OPERATION_DISABLED` error after argument normalization (so JSON-string serialization from VS Code clients is handled),
  3. `executeTool` carries its own guard so direct callers (not just the MCP dispatcher) are covered.

  The tests explicitly exercise the middle and inner layers (`testExecuteTool` paths in section 2 and 3), which is the harder half to get right.
- **No mutation of the shared `n8nManagementTools` array.** `JSON.parse(JSON.stringify(original))` is the right deep-clone choice for pure-data tool defs, and the test at line 742–752 pins this explicitly — a regression here would be silent and catastrophic, so having the assert is excellent.
- **Caching is done at the right layer.** `getDisabledToolOperations()` parses once and `buildFilteredToolDefinitions()` is invoked inside the same first-call path, so every subsequent `ListTools` request is a `Map.get` lookup. No per-request cloning cost.
- **Safety limits mirror `DISABLED_TOOLS`** (10 KB cap, 50-entry cap) — consistent with the existing precedent and tested directly (`should enforce 50-entry limit`, `should enforce 10KB size limit on env var`).
- **`TOOL_OPERATION_PARAM` is the right abstraction.** Co-locating it with the tool definitions and keeping it small (two entries) means adding a future eligible tool is a single-file, single-line change. The map is necessary because `n8n_executions` uses `action` and `n8n_workflow_versions` uses `mode` — good call surfacing that rather than papering over it with a string-match convention.
- **Interaction with `DISABLED_TOOLS` is tested both ways.** Tool-level block takes precedence when both are set on the same tool (`should block at tool level when tool is in DISABLED_TOOLS`), and they can also target different tools independently — exactly the semantics an operator would expect.
- **Docs are coherent across layers.** `.env.example`, `README.md`, and `HTTP_DEPLOYMENT.md` all agree on the recipe, and `getToolDocumentation` / `getToolsOverview` inject a `Server policy` notice so the model sees the same truth as the dispatcher enforces. The framing in `HTTP_DEPLOYMENT.md` — *"The MCP layer is a convenience knob; the n8n API RBAC is the authoritative enforcement boundary"* — is the correct thing to be telling users, and should stay in.

## Nits (optional, can be follow-ups)

1. **Case sensitivity on operation names.** `disabledOpsForTool.has(String(requestedOp))` is strict-equality, so `action: "DELETE"` would slip past a `n8n_executions:delete` rule. The tool schemas use lowercase enum values and well-behaved clients will match, but a lowercase normalization (`disabledOpsForTool.has(String(requestedOp).toLowerCase())`, with the parser lowering the env var entries too) would close a low-probability bypass. Not a blocker.

2. **Drive-by change on existing `DISABLED_TOOLS` path at line 245–246.** The PR also flips the existing tool-disabled response to include `isError: true`. I agree this is the right MCP shape, but it's worth calling out explicitly in the PR description so reviewers don't miss it — it's a behaviour change for existing users of `DISABLED_TOOLS`, however minor.

3. **All-operations-disabled edge case.** If someone writes `DISABLED_TOOL_OPERATIONS=n8n_executions:get,list,delete`, the resulting enum is `[]`. The tool still appears in `ListTools` but the LLM can't legally call any operation. A `logger.warn` when the filtered enum ends up empty (or auto-promoting the tool into the `DISABLED_TOOLS` set) would be a friendlier failure mode for whoever misconfigures this. Happy for this to land as a follow-up.

None of those block merge in my view.

## Test plan executed (read-only review)

- [x] Read full diff of `src/mcp/server.ts` — verified both dispatch guards and the lazy cache build
- [x] Verified `buildFilteredToolDefinitions` uses deep-clone and the test pins non-mutation
- [x] Verified `TOOL_OPERATION_PARAM` covers both eligible tools with the correct arg names (`action` / `mode`)
- [x] Verified tool-level `DISABLED_TOOLS` check runs **before** the operation-level check in both `CallToolRequestSchema` and `executeTool`
- [x] Read all 383 lines of `tests/unit/mcp/disabled-tool-operations.test.ts` — parser, dispatch for both tools, interaction with `DISABLED_TOOLS`, schema non-mutation, docs notice — coverage is comprehensive
- [ ] Not executed: `npm test -- disabled-tool-operations` (review-only, no local build environment)

Approving. Closes #714 cleanly.

mahmoudnaif:
Hey @diborque2733, Thanks for the detailed review and the approval — went ahead and addressed all three nits before this goes in.

---

**1. Case sensitivity**

Applied lowercase normalisation at both ends:
- **Parse time** — `ops.forEach(op => existing.add(op.toLowerCase()))` in `getDisabledToolOperations()`, so the stored `Set` is always lowercase regardless of how the operator writes the env var.
- **Comparison time** — `disabledOpsForTool.has(String(requestedOp).toLowerCase())` in both the `CallToolRequestSchema` guard and the `executeTool` defense-in-depth guard, so a client sending `action: "DELETE"` is correctly blocked by an `n8n_executions:delete` rule.

Two new tests added: one verifying the parser normalises uppercase env var entries, one verifying the dispatch guard catches uppercase client-sent values.

---

**2. `isError: true` on the `DISABLED_TOOLS` path**

Added a highlighted **Note (behaviour change)** callout under the `CallToolRequestSchema` bullet in the PR description — should be visible to anyone scanning the diff.

---

**3. All-operations-disabled edge case**

Went with the `logger.warn` approach. The alternative (auto-promoting the tool into `DISABLED_TOOLS`) looked clean on paper but required coordinating two independent lazy caches — `disabledToolsCache` is already populated and used to filter the tool list *before* `getDisabledToolOperations()` runs, so the first `ListTools` response would still expose the empty-enum tool regardless. The warn is purely additive, fires at the right moment, and gives the operator a direct pointer to the fix. Any call attempt still fails with `OPERATION_DISABLED` so there is no functional gap.

One new test added asserting `logger.warn` is called with the correct message when the entire enum is filtered out.

---

Test suite: **33 tests, all passing**.

czlonkowski:
Thanks @mahmoudnaif — this is a well-built feature. The per-operation gating (deep-cloned tool defs, enum stripping at ListTools, defense-in-depth at both dispatch points, the 10KB/50-entry limits mirroring DISABLED_TOOLS) is solid, and the later commits already handled case-insensitivity nicely.

I rebased it onto current `main` (it was ~109 commits behind and conflicting) and pushed a few follow-ups directly to the branch:

1. **Rebase / conflict resolution.** The main conflict was in `src/mcp/server.ts` around the tool-dispatch logging, which `main` reworked for GHSA-wg4g-395p-mqv3 (it now logs redacted metadata via `summarizeToolCallArgs`). I kept your relocated operation guard (after arg normalization, with `isError: true`) and dropped the older raw-args logging so the security fix is preserved.
2. **Removed the `truncate` mode** (blocking). `n8n_workflow_versions`' real `mode` enum is `list/get/rollback/delete/prune` — there is no `truncate`, so that recipe entry was a silent no-op that misled operators. Removed it from `.env.example`, README, `HTTP_DEPLOYMENT.md`, and the tests.
3. **Startup validation warning.** Configured entries that can never match (unknown tool name, or an operation not in the tool's enum — e.g. `n8n_execution:delete` or `n8n_executions:remove`) now emit a `logger.warn` at parse time, so typos are visible rather than silently leaving an operation enabled. For a security-oriented feature, a silent no-op is the dangerous failure mode.

`npm run build`, `npm run typecheck`, and the full `tests/unit/mcp` suite (738 tests, incl. the 34 disabled-tool-operations tests) are green. Recommending merge once CI confirms.

czlonkowski:
Follow-up from a Codex review pass: one more refinement that directly serves this feature's read-only-deployment purpose.

`buildFilteredToolDefinitions` stripped disabled operations from the enum but left the cloned tool's MCP **annotations** untouched. So a read-only deployment that disables every destructive op (e.g. `n8n_executions:delete` or `n8n_workflow_versions:delete,rollback,prune`) still advertised `readOnlyHint: false` / `destructiveHint: true`. Hosts that honor those hints could keep gating/hiding the tool even though only read paths (`list`/`get`) remain — defeating the point.

Fix pushed to the branch: added `DESTRUCTIVE_TOOL_OPERATIONS` and, when no destructive value remains in the filtered enum, set `readOnlyHint: true` / `destructiveHint: false` on the clone. A *partial* disable that still leaves a destructive op keeps `destructiveHint: true`. Two tests added (full-disable → read-only annotations; partial-disable → stays destructive). `npm run build`, `npm run typecheck`, and the full `tests/unit/mcp` suite (740 passing, incl. 36 disabled-tool-operations tests) are green.

The enforcement path itself was already correct — this just keeps the advertised metadata consistent with the filtered surface.

Issue #611, opened 2026-02-25
`undefined` value not accepted in MCP tool calls from VS Cod


michox:
## Summary

The `n8n_update_partial_workflow` tool documents using `undefined` as a sentinel value to remove node properties (e.g., migrating from deprecated `continueOnFail` to `onError`). However, `undefined` is not a valid JSON value, making this feature **impossible to use from any MCP client** that communicates via JSON where the MCP call layer enforces valid json — observed in VS Code + GitHub Copilot Chat.

## Steps to Reproduce

1. Deploy a workflow with a node that has `continueOnFail: true`
2. Attempt to migrate to `onError: "continueRegularOutput"` with a Github Copilot Agent using the documented pattern:

```javascript
n8n_update_partial_workflow({
  id: "Gv9r9ZGukFkgrBjT",
  operations: [{
    type: "updateNode",
    nodeName: "Mark_As_Read",
    updates: {
      continueOnFail: undefined,   // ← Cannot be expressed in JSON
      onError: "continueRegularOutput"
    }
  }]
});
```

3. Attempting to pass `undefined` directly and viable substitutes:

| # | Attempt | Value Sent | Result | Verified |
|---|---------|-----------|--------|----------|
| 1 | `undefined` (literal) | `continueOnFail: undefined` | ❌ **Framework parse failure** — entire `operations` parameter rejected with `"must be array"`. The value is not valid JSON, so the tool call never reaches the MCP server. | ✅ Tested |
| 2 | `null` | `continueOnFail: null` | ❌ Server rejects: `"Expected boolean, received null"` | ✅ Tested |
| 3 | `"__UNDEFINED__"` | `continueOnFail: "__UNDEFINED__"` | ❌ Server rejects: `"Expected boolean, received string"` | ✅ Tested |
| 4 | `false` | `continueOnFail: false` | ⚠️ Accepted, but property remains on the node → subsequent validation error: `"Cannot use both continueOnFail and onError"` | ✅ Tested |
| 5 | omit key | `{ onError: "continueRegularOutput" }` | ⚠️ Only sets `onError`, does not remove existing `continueOnFail` → same mutual exclusivity error | ⚠️ Inferred (updateNode merges, so omitting a key means "don't change") |

**All paths fail.** The only workaround is to delete and recreate the entire workflow.

## Expected Behavior

There should be a JSON-compatible sentinel value that triggers property removal in `updateNode` operations.

## Root Cause

[...]

Issue #600, opened 2026-02-18
n8n_update_partial_workflow: operations array received as string from Claude Code (VS Code extension)


YerayRodri:
## Bug Description

When calling `n8n_update_partial_workflow` from Claude Code (VS Code extension), the `operations` parameter arrives at the MCP server as a **string** instead of an **array**, causing Zod validation to fail.

## Error Received

```json
{
  "success": false,
  "error": "Invalid input",
  "details": {
    "errors": [
      {
        "code": "invalid_type",
        "expected": "array",
        "received": "string",
        "path": ["operations"],
        "message": "Expected array, received string"
      }
    ]
  }
}
```

## Environment

- **MCP Server version**: 2.35.2 (Docker image 2026-02-10)
- **Docker image**: `ghcr.io/czlonkowski/n8n-mcp:latest`
- **Client**: Claude Code (VS Code extension)
- **MCP Mode**: stdio
- **Platform**: macOS Darwin 25.2.0

## Steps to Reproduce

Call `n8n_update_partial_workflow` with any operation array from Claude Code:

```json
{
  "id": "wf_id",
  "intent": "Rename workflow",
  "operations": [{"type": "updateName", "name": "New Name"}]
}
```

## Expected Behavior

The `operations` parameter is parsed as an array and the operation is applied successfully.

## Actual Behavior

The MCP server receives `operations` as a JSON string instead of a native array and Zod rejects it with `invalid_type`.

## Possible Fix

Add a Zod preprocessor to coerce JSON strings to arrays before validation:

```typescript
operations: z.preprocess(
  (val) => (typeof val === "string" ? JSON.parse(val) : val),
  z.array(operationSchema)
)
```

## Notes

All other tools with simple string parameters work correctly (`n8n_list_workflows`, `n8n_get_workflow`, `n8n_autofix_workflow`, `n8n_health_check`). Only tools requiring array parameters are affected.

czlonkowski:
Thanks for the clear report and the spot-on suggested fix, @YerayRodri — a Zod preprocessor that coerces a JSON string into the array before validation is exactly what shipped.

This is now resolved in current releases. The MCP input normalizer (`src/utils/mcp-input-normalizer.ts`, `normalizeMcpJsonValue`) JSON-parses stringified arguments before Zod validation, and the `n8n_update_partial_workflow` schema wraps the field accordingly:

```ts
// src/mcp/handlers-workflow-diff.ts
[...]

PR #727, opened 2026-04-14
fix: parse JSON-stringified workflow params in MCP schemas (v2.47.8, #611)

czlonkowski:
## Summary

Fixes the second bug reported in #611 (by @Mte90): `n8n_create_workflow` fails with `"nodes must be an array, got string"` on workflows with 3+ nodes or complex nested parameters (resource-locator `__rl` objects, filters, etc.). 1-2 node payloads happened to slip through, which made it look intermittent.

**Root cause:** VS Code + GitHub Copilot (and some other MCP clients) serialize array/object tool arguments as JSON strings rather than native JSON types. `n8n_update_partial_workflow` already preprocessed its `operations` parameter with `tryParseJson` (from the prior #600/#611 fix in 9133f5e), but `n8n_create_workflow`, `n8n_update_full_workflow`, and `n8n_list_workflows` did not.

**Note on #611's original report:** The `undefined`-as-property-removal-sentinel problem raised by @michox is already resolved in the current codebase — `setNestedProperty` in `workflow-diff-engine.ts` accepts `null` as the deletion sentinel (commit 65ab94d), and 9133f5e added the JSON round-trip that strips `undefined` from tool args. This PR addresses only the new regression surfaced by @Mte90.

## Changes

- `src/mcp/handlers-n8n-manager.ts` — Wrap `nodes`, `connections`, and `settings` in `createWorkflowSchema` and `updateWorkflowSchema` with `z.preprocess(tryParseJson, ...)`. Same for `tags` in `listWorkflowsSchema`.
- `src/mcp/handlers-n8n-manager.ts` — Relocate `tryParseJson` from the bottom of the file (line ~2761) to next to its first usage, since it's now consumed at the top. Export is preserved (imported by `handlers-workflow-diff.ts`).
- `src/mcp/server.ts` — Both silent `catch {}` blocks in `coerceStringifiedJsonParams` now emit `logger.warn` with the parse error, a 200-char value preview, and the value length. Previously these failures were invisible until downstream Zod errors surfaced.

## Verification

- All 5333 unit/integration tests pass (`npm test`, 196 files, 38.97s).
- Live MCP regression pass via `n8n-mcp-tester` against the testing instance:
  - Created a 4-node workflow (Webhook → Filter → Set → RespondToWebhook) with nested filter conditions, assignment arrays, and a `settings` object — the exact shape that was failing. Created, fetched, validated, updated (both full and partial), and deleted successfully.
  - `n8n_health_check`, `n8n_list_workflows`, `search_nodes`, `get_node`, `n8n_validate_workflow`, `n8n_update_partial_workflow` (updateNode) all pass with no regressions.

## Test plan

- [x] `npm run build`
- [x] `npm test` (5333 pass, 74 skipped)
- [x] Live MCP integration test with `n8n-mcp-tester` against testing instance
- [x] Manual verification: 4-node workflow with complex nested params creates successfully


Mte90:
Cool :-D

czlonkowski:
@Mte90 should be live in some 15 minutes when npm is published :)

PR #730, opened 2026-04-15
fix(validator): accept columns resourceMapper for Google Sheets read/update

infobewaize:
## Problem

`validateGoogleSheetsUpdate` and `validateGoogleSheetsRead` raise false-positive `missing_required` errors on valid Google Sheets v4+ configurations that use the `columns` resourceMapper (`mappingMode: "defineBelow"` / `"autoMapInputData"`) instead of the legacy `range` + `values` fields.

The `columns` mapper carries both the address (via `matchingColumns` + schema) and the data (via `columns.value`), so `range` and `values` are ignored at runtime. The `Append` validator already has the right shape — accepting either `range` OR `columns` — but Read and Update were never updated to match.

## Reproducer

Minimal config that currently fails validation but works fine at runtime:

```json
{
  "operation": "update",
  "documentId": {"__rl": true, "value": "...", "mode": "list"},
  "sheetName": {"__rl": true, "value": "...", "mode": "list"},
  "columns": {
    "mappingMode": "defineBelow",
    "value": {"Email": "={{ $json.email }}", "Status": "active"},
    "matchingColumns": ["Email"]
  }
}
```

Produces:
- `Range is required for update operation`
- `Values are required for update operation`

## Fix

Mirrors the existing `validateGoogleSheetsAppend` logic:

- `validateGoogleSheetsRead`: accept `columns` as alternative to `range`
- `validateGoogleSheetsUpdate`: accept `columns.mappingMode` / `columns.value` as alternative to both `range` and `values`
- No behavior change for legacy `range`/`values` configs — they validate exactly as before

## Tests

- Updated existing error-message assertions for read/update
- Added positive tests confirming `columns.mappingMode: "defineBelow"` is accepted for both operations

> Note: local test suite couldn't run (Chromebook ARM, `/tmp` out of space during `npm install`). CI will validate. The change is small and symmetric with existing Append semantics.

## Checklist

- [x] "Allow edits by maintainers" is enabled on this PR
- [x] Existing test assertions updated to match new error messages
- [x] Added positive coverage for the new accepted config shape

Co-authored-by: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/services/node-specific-validators.ts
+++ b/src/services/node-specific-validators.ts
@@ -310,12 +310,22 @@ export class NodeSpecificValidators {
     errors.push(...filteredErrors);
   }
   
+  /**
+   * In Google Sheets v4+, the `columns` resourceMapper (mappingMode "defineBelow" /
+   * "autoMapInputData") carries the range/values via matchingColumns + schema, so the
+   * legacy range/values fields are not required. An empty `columns: {}` object does not
+   * count — a real mapping has at least a mappingMode or a value.
+   */
+  private static hasColumnsMapping(config: any): boolean {
+    return !!(config.columns && (config.columns.mappingMode || config.columns.value));
+  }
+
   private static validateGoogleSheetsAppend(context: NodeValidationContext): void {
     const { config, errors, warnings, autofix } = context;
 
     // In Google Sheets v4+, range is only required if NOT using the columns resourceMapper
     // The columns parameter is a resourceMapper introduced in v4 that handles range automatically
-    if (!config.range && !config.columns) {
+    if (!config.range && !this.hasColumnsMapping(config)) {
       errors.push({
         type: 'missing_required',
         property: 'range',
@@ -341,7 +351,7 @@ export class NodeSpecificValidators {
 
     // In Google Sheets v4+, range is only required if NOT using the columns resourceMapper
     // (same semantics as append operation)
-    if (!config.range && !config.columns) {
+    if (!config.range && !this.hasColumnsMapping(config)) {
       errors.push({
         type: 'missing_required',
         property: 'range',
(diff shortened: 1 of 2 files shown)


czlonkowski:
Thanks @infobewaize — this is a correct and useful fix. The false-positive `missing_required` on v4+ Google Sheets configs that use the `columns` resourceMapper was a real pain point.

I've pushed two follow-up changes directly to the branch and rebased it onto current `main`:

1. **Fixed a failing test.** A leftover assertion in the untouched `common validations` block (`should require range for read operation`) still expected the old `"Range is required for read operation"` message and would have turned CI red after your message change. It now asserts the new `"Range or columns mapping is required..."` message and additionally checks that `sheetId` is not flagged (it comes from credentials) — keeping that test's original intent.
2. **Removed a Read-vs-Update asymmetry.** Read/Append used a plain `!config.columns` check while Update used a stricter `hasColumnsMapping` (requiring `mappingMode`/`value`). I factored a shared `hasColumnsMapping()` helper and use it in all three, so an empty `columns: {}` no longer satisfies the requirement consistently.

`npm run typecheck` and the full `node-specific-validators` suite (189 passing) are green on the rebased branch. Recommending merge once CI confirms.

czlonkowski:
Follow-up from a Codex review pass, verified against the node schema: I've adjusted the **read** portion of this PR.

The `columns` resourceMapper only exists for the **write** operations — in the Google Sheets node schema `columns` is shown for `append`, `update`, and `appendOrUpdate` (resource `sheet`) only. The **read** operation has no `columns` parameter (its params are `sheetName`, `filtersUI`, `combineFilters`, `options`). So accepting `columns` as an alternative to `range` for read meant a read config with a stray `columns` object would bypass the location check, while n8n ignores `columns` for reads entirely.

Change pushed to the branch:
- `validateGoogleSheetsRead` reverts to requiring `range` (its prior behavior).
- The columns handling is **kept for append and update**, where the resourceMapper genuinely exists — that's the correct, valuable part of this PR.
- Read tests updated to match (read still requires `range` even when a `columns` object is present).

Net: this PR now correctly fixes the false-positive for **update** (and append) v4 configs using the `columns` resourceMapper, without introducing a read-side false negative. `npm run typecheck` and the full `node-specific-validators` suite (189 passing) are green.

Note (not changed): for `update`, `hasColumnsMapping` treats `mappingMode` alone as sufficient. That's deliberate — requiring a non-null `value` would re-introduce false positives for `autoMapInputData` and expression-based mappings — but flagging it in case you'd prefer stricter behavior.

Issue #659, opened 2026-03-23
addConnection creates broken connections with "type": "0" instead of "type": "main"


VolodX:
## Bug Description

`addConnection` operation in `n8n_update_partial_workflow` creates connections with `"type": "0"` instead of `"type": "main"` when `sourceOutput: 0` is specified for regular (non-AI) nodes. This results in broken connections in n8n — the target node shows "No input connected" and `$json` is empty.

## Steps to Reproduce

```javascript
n8n_update_partial_workflow({
  id: "workflow_id",
  operations: [
    {
      type: "addConnection",
      source: "IF Node",
      target: "HTTP Request",
      sourceOutput: 0,
      targetInput: 0
    }
  ]
})
```

## Expected Result

Connection with `"type": "main"`:
```json
{
  "IF Node": {
    "main": [
      [{ "node": "HTTP Request", "type": "main", "index": 0 }]
    ]
  }
}
```

## Actual Result

Connection with `"type": "0"`:
```json
{
  "IF Node": {
    "main": [
      [{ "node": "HTTP Request", "type": "0", "index": 0 }]
    ]
  }
}
```

The n8n UI shows the connection visually, but the target node displays "No input connected". Data does not flow through properly — `$json` references return empty/undefined values.

## Impact

- Connections appear valid in UI but data doesn't flow
- Downstream nodes fail silently (empty `$json`)
- Hard to diagnose — the connection looks correct visually
- In our case, a live production workflow was broken affecting multiple customer conversations

## Workaround

1. For IF nodes: use `branch: "true"` / `branch: "false"` instead of `sourceOutput: 0`
2. For regular nodes: omit `sourceOutput` entirely (let it default)
3. Manual fix: delete and recreate the connection in n8n UI

## Suggested Fix

When `sourceOutput` is a number (not a string like `"ai_tool"`), the connection `type` field should default to `"main"`, not the stringified number.

```javascript
// Pseudocode
if (typeof sourceOutput === 'number') {
  connectionType = 'main';
  outputIndex = sourceOutput;
} else {
  connectionType = sourceOutput; // "ai_tool", "ai_languageModel", etc.
}
```

## Environment

- n8n: v2.2.5 (Self Hosted)
- n8n-mcp: latest (as of 2026-03-23)

czlonkowski:
[...]

Issue #537, opened 2026-01-15
addConnection operation creates malformed connection types for multi-output nodes (IF, Switch)


bvanorsdel:
## Description

When using `n8n_update_partial_workflow` with `addConnection` operations targeting IF node outputs, the resulting connections use invalid type values that cause rendering/deletion issues in the n8n UI.

## Environment

- n8n-MCP version: 2.33.2
- n8n instance: Cloud (riveting.app.n8n.cloud)
- Node type affected: `n8n-nodes-base.if` (typeVersion 2.2)

## Steps to Reproduce

1. Create or identify a workflow with an IF node
2. Use `n8n_update_partial_workflow` to add connections from IF node outputs:
```json
{
  "id": "workflow-id",
  "operations": [
    {
      "type": "addConnection",
      "source": "My IF Node",
      "target": "Target Node A",
      "sourceOutput": "0",
      "targetInput": "0"
    },
    {
      "type": "addConnection",
      "source": "My IF Node",
      "target": "Target Node B",
      "sourceOutput": "1",
      "targetInput": "0"
    }
  ]
}
```

## Expected Result
```json
{
  "My IF Node": {
    "main": [
      [{ "node": "Target Node A", "type": "main", "index": 0 }],
      [{ "node": "Target Node B", "type": "main", "index": 0 }]
    ]
  }
}
```

## Actual Result
```json
{
  "My IF Node": {
    "0": [
      [{ "node": "Target Node A", "type": "0", "index": 0 }]
    ],
    "1": [
      [{ "node": "Target Node B", "type": "1", "index": 0 }]
    ]
  }
}
```

## Impact

- Connections appear visually in n8n canvas but behave incorrectly
- Clicking the trash icon on the connector does nothing (can't delete)
- Workflow execution may fail or route incorrectly
- Only fixable via `n8n_update_full_workflow` with manually corrected connection structure

## Workaround

Use `n8n_update_full_workflow` instead, passing the complete connections object with correct `"main"` structure.

## Root Cause Hypothesis

The `addConnection` operation appears to use the `sourceOutput` parameter value ("0", "1") directly as both the connection key and the type value, rather than:

1. Using "main" as the parent key
2. Using the sourceOutput as the array index within `main`
3. Always setting `type: "main"` in the connection object

curioushacker11:
Hey, have you found a solution to this issue?

PR #750, opened 2026-04-27
fix(diff-engine): scope addConnection duplicate check to sourceIndex (#738)

### LINKED ISSUE #738: addConnection throws "Connection already exists" when connecting multiple Switch outputs to same target node
## Bug: `addConnection` uniqueness check ignores `sourceIndex`

**n8n-mcp version:** 2.47.4

### Steps to reproduce

1. Have a Switch node with many outputs (e.g. 20+ outputs)
2. A target node already has connections from **some** Switch outputs
3. Use `n8n_update_partial_workflow` with `addConnection` to connect **additional** Switch outputs (different `sourceIndex` or `case` values) to the **same** target node

### Expected

Each connection added successfully — `(source, sourceIndex, target)` is a unique triple, so no conflict.

### Actual

```
Connection already exists
```

The tool checks uniqueness by `(source, target)` pair only, ignoring `sourceIndex`. Any second connection to the same target node — regardless of which output index it comes from — is rejected.

### Workaround

Direct n8n REST API `PUT /api/v1/workflows/{id}` with the full workflow JSON patched manually.

### Impact

Cannot route multiple Switch branches to a shared fallback node (e.g. a Manual Review / error handler node) using n8n-mcp tools. Forces direct REST API usage for a standard workflow pattern.

### Context

Encountered while building a 20+ output Switch node where several branches share a common fallback DB node. Both `sourceIndex` and `case` parameters were tried — same error in both cases.

SAY-5:
Closes #738.

## What

`validateAddConnection` checked uniqueness on `(source, sourceOutput,
target)` only, scanning the entire `existing: Connection[][]` array
across **all** `sourceIndex` slots. Two outputs of a Switch node fanning
into the same target node — e.g. several branches sharing a
fallback / error handler, exactly the scenario in the bug report — got
rejected with `Connection already exists` even though
`(source, sourceIndex=N, target)` wasn't actually a duplicate.

## Fix

`validateAddConnection` now:

- resolves `sourceIndex` via `resolveSmartParameters`, the same helper
  `applyAddConnection` and rewire already use (it honours `sourceIndex`,
  `case=`, `branch=`, and the numeric-sourceOutput remap from
  #537 / #659);
- only checks the `existing[sourceIndex]` slot for the target;
- includes the offending `sourceIndex` in the error message so callers
  can distinguish a genuine duplicate from an unrelated index.

## Tests

`tests/unit/services/workflow-diff-engine.test.ts`:

- **New** `should allow connecting different sourceIndex outputs to the
  same target (#738)`: pre-wires `Slack -> HTTP Request` at
  `sourceIndex=0`; `addConnection` at `sourceIndex=1` is accepted; the
  resulting workflow has both indices populated.
- **New** `should still reject duplicate connections at the same
  sourceIndex (#738)`: re-adding the same edge at the same index still
  errors with `Connection already exists` and now includes
  `sourceIndex=0` in the message.
- The pre-existing `should reject duplicate connections` continues to
  pass — the default-sourceIndex-0 path is unchanged.

## Verification

- `npx vitest run tests/unit/services/workflow-diff-engine.test.ts -t addConnection`
  → 7 passing tests.
- I confirmed locally that reverting just the validation block back to
  the old sourceIndex-blind code makes **both** new #738 tests fail —
  i.e. the new tests catch the regression cleanly.

### DIFF
--- a/src/services/workflow-diff-engine.ts
+++ b/src/services/workflow-diff-engine.ts
@@ -697,15 +697,20 @@ export class WorkflowDiffEngine {
       return `Target node not found: "${operation.target}". Available nodes: ${availableNodes}. Tip: Use node ID for names with special characters (apostrophes, quotes).`;
     }
 
-    // Check if connection already exists
-    const sourceOutput = operation.sourceOutput || 'main';
+    // Check if a connection already exists *at the same sourceIndex*.
+    // For multi-output nodes (Switch, IF, etc.) the connection is keyed
+    // by (source, sourceOutput, sourceIndex, target) — two outputs of
+    // the same Switch can both legitimately fan into the same target
+    // node (e.g. several branches sharing a fallback / error handler).
+    // Without scoping the check to sourceIndex, the second branch was
+    // rejected with "Connection already exists" even though
+    // (source, sourceIndex=N, target) didn't actually exist yet (#738).
+    const { sourceOutput, sourceIndex } = this.resolveSmartParameters(workflow, operation);
     const existing = workflow.connections[sourceNode.name]?.[sourceOutput];
-    if (existing) {
-      const hasConnection = existing.some(connections =>
-        connections.some(c => c.node === targetNode.name)
-      );
+    if (existing && Array.isArray(existing[sourceIndex])) {
+      const hasConnection = existing[sourceIndex].some(c => c.node === targetNode.name);
       if (hasConnection) {
-        return `Connection already exists from "${sourceNode.name}" to "${targetNode.name}"`;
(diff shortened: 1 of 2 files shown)


SAY-5:
Closing — the sourceIndex-scoped duplicate check has since landed in main (`src/services/workflow-diff-engine.ts:707-720`) along with a *Switch / multi-output → shared target (Issue #738)* test suite (`tests/unit/services/workflow-diff-engine.test.ts:1283`). The behavior change and regression coverage from this PR are both already in. No-op merge from this PR.

Issue #744, opened 2026-04-23
Bug: n8n_update_partial_workflow validateOnly=true passes, but validateOnly=false fails with Zod "_zod" / invalid record key errors (v2.47.14)


Valirius:
Title
Bug: n8n_update_partial_workflow validateOnly=true passes, but validateOnly=false fails with Zod "_zod" / invalid record key errors (v2.47.14)
Body
## Summary
`n8n_update_partial_workflow` can validate operations successfully (`validateOnly=true`), but fails to save the same operations (`validateOnly=false`) with structural validation errors.
This blocks even metadata-only updates (e.g. `updateName`) on otherwise healthy workflows.
---
## Environment
- n8n-mcp version: `2.47.14`
- n8n API URL: `https://n8n.contentceh.ru/`
- Connectivity checks:
  - `n8n_health_check`: OK
  - `n8n_list_workflows`: OK
- Affected workflows: multiple (W-02/W-03/W-04/W-05 family)
---
### 1) Validate-only succeeds
`n8n_update_partial_workflow`
```json
{
  "id": "q7xVY8nLm2R5pCkD",
  "validateOnly": true,
  "continueOnError": false,
  "operations": [
    { "type": "updateName", "name": "W-05b RAG Query Surface" }
  ]
}
```
Result: success=true, valid=true.
2) Apply fails for same operation

n8n_update_partial_workflow
```json
{
  "id": "q7xVY8nLm2R5pCkD",
  "validateOnly": false,
  "continueOnError": false,
  "operations": [
    { "type": "updateName", "name": "W-05b RAG Query Surface" }
  ]
}
```
Result: success=false, saved=false.
---
Actual Error
Errors include:
- Cannot read properties of undefined (reading '_zod') (on multiple node indexes)
- Invalid key in record
- Invalid input: expected object, received string
- Error paths look like node names, e.g.:
  - ["W05b Webhook Trigger"]
  - ["W05b Set Context"]
  - etc.
---
Expected Behavior
- Metadata-only operations like updateName should save successfully.
- validateOnly=true and validateOnly=false should run against the same canonical workflow shape and report consistent structural results.
---
Root Cause (from local debugging)
There appear to be two related issues:
1. Zod record schema misuse in structural validation
   In n8n-validation, record schemas are defined with single-argument z.record(...) forms (e.g. z.record(z.unknown()), z.record(z.object(...))), which in current Zod runtime leads to key/value schema mismatch and invalid_key errors.
[...]

PR #752, opened 2026-04-28
fix: declare workflow record key schemas

Valirius:
## Summary
- Use explicit string-keyed Zod records for workflow node `parameters` and `credentials`.
- Use an explicit string-keyed Zod record for workflow connection maps.
- Keeps create/validate workflow paths consistent across supported Zod versions.

## Verification
- `npx -y -p typescript@5.8.3 tsc -p tsconfig.build.json --pretty false`


czlonkowski:
Thank you @Valirius for spotting this and for the clean, well-explained fix — the Zod 3/4 single-arg `z.record` reinterpretation was a real footgun.

We're going to close this one as already resolved. The same change is now in `main`: `src/services/n8n-validation.ts` declares `parameters: z.record(z.string(), z.unknown())` and `credentials: z.record(z.string(), z.unknown()).optional()` (lines 23-24), and `workflowConnectionSchema` passes the explicit `z.string()` key schema (lines 52-53) — with a comment referencing issue #744 and the exact Zod 3/4 rationale you described. This landed via #757 ("validateOnly path consistency + Zod 4 record-key compatibility (#744)"), which addressed the same root cause together with a related validateOnly path fix, and issue #744 is now closed.

Because the branch is also ~100 commits behind and now conflicts on this file, there's nothing left to merge here. Really appreciate the contribution and the diagnosis — please keep them coming!

PR #753, opened 2026-04-28
fix: validate partial workflow previews

Valirius:
## Summary
- Return the canonical post-diff workflow from `validateOnly` partial update runs.
- Run structural workflow validation before returning successful preview responses, matching the real save path.
- Add a regression script covering connection preservation and validateOnly/apply shape parity.

## Verification
- `npx -y -p typescript@5.8.3 tsc -p tsconfig.build.json --pretty false`
- `node dist/scripts/test-update-partial-save-path-regression.js`


czlonkowski:
Thank you, @Valirius (and @Contentceh), for this fix and for the thorough regression coverage — really appreciated.

Closing this as already resolved: the same validateOnly consistency bug (issue #744) was fixed in main by #757 (commit 7fc714d), which landed on 2026-04-28. Both parts of your change are now present:

- `src/mcp/handlers-workflow-diff.ts` runs `validateWorkflowStructure` before the validateOnly early-return and reports `data.valid: structureErrors.length === 0`, so the preview and apply paths agree (see lines ~251-282).
- `src/services/workflow-diff-engine.ts` returns `workflow: workflowCopy` from both validateOnly return sites (atomic + continueOnError modes), so the handler can validate the post-diff snapshot.

The merged version also added suite-integrated unit tests in `tests/unit/mcp/handlers-workflow-diff.test.ts` (including the #744 valid=false case) and `tests/unit/services/workflow-diff-engine.test.ts`, covering the same scenarios as your regression script.

The branch is now ~101 commits behind main and conflicts on these rewritten files, so there's nothing left to merge here. Thanks again for diagnosing the same root cause independently — please keep the contributions coming.

PR #766, opened 2026-05-02
feat: local LLM template metadata + refresh template store (v2.50.0)

czlonkowski:
## Summary

Three things land together (all on `chore/update-templates-may-2026`):

1. **Refresh template store** from n8n.io (was stale since 2025-12-24): 2,352 templates, 156 ranked node configurations, 2,351/2,352 with regenerated metadata.
2. **Refresh community node store**: 830 nodes (was 768, +62 new) from the Strapi verified list and top-100 npm packages. 825/830 carry npm READMEs and AI documentation summaries (99.4% coverage; the 5 misses publish no README on npmjs, so there is no source text).
3. **New: local LLM support for template metadata.** When `N8N_MCP_LLM_BASE_URL` is set, `fetch:templates --metadata-only` routes through any OpenAI-compatible server (vLLM, Ollama, llama.cpp `/v1`). Falls back to OpenAI Batch API otherwise. New `SequentialMetadataProcessor` for direct (non-batch) generation since vLLM doesn't implement `/v1/batches`. Tightened metadata system prompt to fix a prompt-leak failure mode on smaller open-source models.

DB totals: **1,650 nodes** (820 base + 830 community), **2,352 templates** with metadata.

## How it ran

- vLLM 0.18.0 + `Qwen/Qwen3.5-9B` on a local RTX 3090, two passes:
  - Templates: concurrency 40, ~13 min, 99.96% success.
  - Community summaries: concurrency 3 (the existing batch processor's default), ~13 min for 244 nodes; 9 stragglers retried at 8k context (8 succeeded), 1 final at 16k context (succeeded).
- Started/stopped vLLM cleanly each round; GPU released between phases.

## New env vars

| var | default | purpose |
|---|---|---|
| `N8N_MCP_LLM_BASE_URL` | _unset_ (uses OpenAI) | OpenAI-compatible `/v1` endpoint |
| `N8N_MCP_LLM_MODEL` | `Qwen/Qwen3.5-9B` | model id sent to the local server |
| `N8N_MCP_LLM_API_KEY` | `EMPTY` | optional bearer for the local server |
| `N8N_MCP_LLM_CONCURRENCY` | `40` | parallel chat completion calls (templates path) |

`generate-community-docs.ts` already supported `N8N_MCP_LLM_BASE_URL`; this PR doesn't change that path.

## Test plan

- [x] `npm run build` clean
- [x] `npm run typecheck` clean
- [x] Smoke test: 20 templates against vLLM Qwen3.5-9B → 0 failures, valid JSON-schema output
- [x] Full template run: 2,332 templates → 2,315 first pass + 16 retry pass → 1 skipped
- [x] Community node fetch: upsert preserves existing READMEs and AI summaries
- [x] Community docs incremental: 235 + 8 + 1 (three passes at increasing context) → 825/830 covered
- [x] Community node count preserved through template fetch (768 mid-run → 830 after community fetch)
- [x] OpenAI Batch path still typechecks (unchanged callable contract)
- [ ] CI pipeline

## Notes

- Pre-fetch DB backup at `/tmp/nodes-pre-template-update-20260502-093230.db` (local only).
- Bump: 2.49.0 → 2.50.0.


czlonkowski:
Addressed Copilot review:

- **`N8N_MCP_LLM_CONCURRENCY` validation** (fetch-templates.ts:402): NaN / 0 / negative values now fall back to 40 with a warning printed instead of silently creating zero workers.
- **Shared prompt body** (metadata-generator.ts:322): `createBatchRequest()` now delegates to `buildChatRequest()`. The leak-resistant system prompt now applies to both the OpenAI Batch path and the local direct path verbatim.

Also in this push:
- Lowered the CI templates-table floor from 2,500 to 2,000 (n8n.io currently lists 2,352; the old floor assumed the previous catalogue size).
- README node/template counts refreshed: 1,650 / 820 / 830, 2,352 templates.

Note on `should optimize rebuilding FTS index`: that test is pre-existing flake — `TestDataGenerator.generateTemplate()` uses `Math.floor(Math.random() * 100000)` for `workflow_id` (test-utils.ts:358), so 100 generated rows hit a birthday-paradox collision often enough to fail intermittently in CI. Tracking separately, not in scope for this PR.

czlonkowski:
Pushed FTS5 fix into this PR (`ae55533`):

`TestDataGenerator.generateTemplate()` used `Math.floor(Math.random() * 100000)` for the `id` field — the same value that becomes `workflow_id` in the templates table. With 100 generated rows the birthday-paradox collision rate is ~5%, which is why `should optimize rebuilding FTS index` flaked with `UNIQUE constraint failed: templates.workflow_id`.

Switched to a monotonic counter starting at 100_000. Verified 5 consecutive clean local runs of `fts5-search.test.ts` (18 passed, 1 skipped each time).

czlonkowski:
Addressed second Copilot review (`e752bca`):

- **Threshold tightened to 2,200** (database-population.test.ts:198): ~6% cushion below the current 2,352 floor. Catches partial-fetch regressions; 2,000 was too loose.
- **URL redaction helper** (fetch-templates.ts:14): `redactUrl()` strips userinfo and query params before logging `N8N_MCP_LLM_BASE_URL`. Both log sites (lines 269, 412) now print just `protocol://host:port/path`.
- **API-key sentinel** (fetch-templates.ts:418): default changed from `EMPTY` to the conventional `not-needed`. Comment notes operators behind a Bearer-validating gateway must set `N8N_MCP_LLM_API_KEY` explicitly.
- **SequentialMetadataProcessor unit tests** (tests/unit/templates/sequential-processor.test.ts): 5 tests covering success path, isolated per-template failure, progress callback, concurrency capping, empty input. OpenAI client is mocked so the suite never touches a network.

PR #769, opened 2026-05-03
fix(workflow-diff): roll back to prior state when n8n PUT fails post-save

pybe:
## What

When `n8n_update_partial_workflow`'s underlying PUT fails (e.g. n8n persisted
the workflow body but the activation step within the same PUT then tripped on
an unsupported `typeVersion`), the workflow was being left in a broken state.
The handler already snapshots the prior workflow into `workflowBefore` for
telemetry, so we piggyback on it: on PUT failure, attempt a rollback PUT to
restore the prior state.

## Why

Reproduced 2026-04-27 against MCP 2.47.14 + a self-hosted n8n. Setting an
unsupported `typeVersion` via `updateNode` returned the misleading
`Cannot read properties of undefined (reading 'execute')` error AND silently
saved the workflow with the broken value, leaving an active error-handler
workflow corrupted until manually reverted.

Concrete repro:

```js
n8n_update_partial_workflow({
  id: "<active-workflow-id>",
  operations: [{
    type: "updateNode",
    nodeId: "<node-id>",
    updates: { typeVersion: 999 }   // any value the running n8n doesn't have
  }]
})
// → MCP returns:
//   { success: false, error: "Invalid request: Cannot read properties of undefined (reading 'execute')", code: "VALIDATION_ERROR", ... }
// → but n8n_get_workflow now shows nodes[].typeVersion: 999 PERSISTED.
//   Next workflow execution will fail.
```

I'll open a separate issue with the full investigation linked to this PR.

## How

Single try/catch around `client.updateWorkflow(...)` in
`src/mcp/handlers-workflow-diff.ts`. Three return paths on failure:

- **Pre-mutation** (no snapshot, e.g. `validateOnly`) — preserve current
  behaviour: rethrow so the existing outer catch handles it
- **Rollback succeeded** — return original error suffixed with
  `(workflow restored to prior state)` + `details.rollbackPerformed = true`
- **Rollback failed** — composite error pointing to `n8n_workflow_versions`
  for manual recovery + `details.rollbackPerformed = false` +
  `details.rollbackError`

Success path is byte-identical to before this PR.

## Tests

Two new unit tests in `tests/unit/mcp/handlers-workflow-diff.test.ts`:
- `should roll back to prior state when n8n PUT fails after persisting body`
  — asserts the rollback PUT is fired with the prior snapshot and the
  response carries the restored-state suffix
- `should report rollback failure when both PUTs fail` — asserts the
  composite error is surfaced with both underlying messages

Existing test `should handle API errors during update` was renamed and updated
because its assertion was the exact behaviour this PR changes. The
`should handle input validation errors` and other tests are untouched.

`tests/unit/mcp/handlers-workflow-diff.test.ts` and
`tests/unit/services/workflow-diff-engine.test.ts`: 41 + N tests pass.

(Note: a few unrelated test files in `tests/unit/mcp/` fail on `main` as well
— `get-node-essentials-examples.test.ts`, `get-node-unified.test.ts` — those
failures predate this PR and are out of scope.)

## Known limitations

1. **Concurrent writers can race the rollback.** If another caller modifies
   the workflow between our GET and our rollback PUT, the rollback could
   clobber their write. Same race window as today's plain PUT, just
   inverted. A future iteration could use the workflow's `versionId` for
   optimistic concurrency.
2. **Rollback fires even when n8n rejected the PUT pre-save** (body never
   persisted). The rollback PUT is then a no-op (re-saves the prior state).
   Cost: one extra HTTP request on the failure path. Chose this over
   guessing n8n's error semantics, which can shift between n8n versions.

## Backwards compatibility

- Success path: byte-identical, no change.
- Failure path: error string gains a suffix (`(workflow restored to prior state)`
  or rollback-failed message); `details.rollbackPerformed` and optionally
  `details.rollbackError` are added. Callers that ignored `details` see only
  the suffix change. Callers that introspect `details` get strictly more info.


pybe:
Closes #770 — full investigation & repro details in the linked issue.

czlonkowski:
Rebased on current main (was at v2.50.0; main is now v2.50.2 after security releases #767, #771) and force-pushed with two refinements on top of @pybe's original commit. Bumped to v2.50.3.

## What I changed

1. **Skip rollback when n8n rejected the PUT before persisting** (addresses limitation #2 from the PR description). After the failed PUT, GET the workflow once and compare `versionId` (with fallback to `versionCounter` then `updatedAt`) to the snapshot. If unchanged → body never persisted → rethrow the original error verbatim, no rollback PUT, no `(workflow restored to prior state)` suffix, `details.rollbackPerformed: false`. The unconditional rollback was misleading the caller in this case.
2. **Surface `details.priorVersionId`** when the snapshot has a `versionId`, so callers can recover via `n8n_workflow_versions` after a rollback failure.
3. **Tri-state version compare with safety net.** If the post-failure GET fails or neither side exposes any version field (older n8n), fall through to attempting rollback — silent corruption is far worse than a redundant PUT.

## Tests added/updated

- Persist-then-fail rollback (existing, updated to assert `priorVersionId`)
- Pre-save rejection (new) — no rollback PUT, no suffix, `rollbackPerformed: false`
- Post-failure GET fails (new) — best-effort rollback fires
- `versionCounter`-only fallback (new) — older n8n still gets rollback, `priorVersionId` correctly omitted
- No version fields on either side (new) — safety net rolls back
- Double-PUT-failure with `priorVersionId` in details (existing, updated)

`tests/unit/mcp/handlers-workflow-diff.test.ts`: 45/45 pass. `tests/unit/mcp`: 611/611. Build and typecheck clean.

## End-to-end validation

Tested live against a real n8n instance with the `typeVersion: 999` repro from #770:

```
error: "Invalid request: Cannot read properties of undefined (reading 'execute') (workflow restored to prior state)"
code: "VALIDATION_ERROR"
details.rollbackPerformed: true
details.priorVersionId: "56a81409-f36c-4a11-b6ce-e4392b043e84"
```

`n8n_get_workflow` after the call confirmed the HTTP Request node was back to its original `typeVersion`. `n8n_workflow_versions list` confirmed the `priorVersionId` is actionable for recovery.

The pre-save rejection branch isn't reachable through the MCP's diff API (the MCP's own pre-PUT validation is tighter than n8n's body validator), so live testing covers persist-then-fail only — the unit tests cover the pre-save branch.

## Out of scope (follow-ups worth filing)

- Optimistic-concurrency rollback via versionId-conditional PUT (limitation #1 in the PR description). Needs n8n PUT semantics research first.
- Telemetry currently records `workflowAfter: workflowBefore` even when rollback fails — minor mislabel that should reflect the persisted-but-broken state instead.

PR #783, opened 2026-05-12
fix(http): preserve shared tenant sessions

LevSky22:
## Summary

Fixes shared multi-tenant session handling so that initializing a new MCP client with the same `x-instance-id` no longer invalidates existing sessions when `MULTI_TENANT_SESSION_STRATEGY=shared`.

Previously, the eager same-instance cleanup ran before checking the configured multi-tenant session strategy. As a result, shared mode still behaved partly like instance mode: a second client initializing for the same tenant removed the first client's session, causing stale `Mcp-Session-Id` / "Session not found or expired" failures.

## Changes

- Gate same-instance eager cleanup behind `MULTI_TENANT_SESSION_STRATEGY=instance`.
- Preserve existing same-instance sessions when the strategy is `shared`.
- Add regression coverage for both behaviors:
  - shared mode keeps same-instance sessions alive
  - instance mode replaces same-instance sessions

## Validation

- `npx vitest run tests/unit/http-server-session-management.test.ts --coverage.enabled=false`
- `npm run build`

## Notes

This keeps the existing `instance` strategy behavior unchanged while making `shared` behave as expected for multiple concurrent MCP clients using the same tenant.


czlonkowski:
Thank you for the fix and the regression tests, @LevSky22! 🙏

We rebased this and #784 onto current `main` (your branch had drifted past several merges, including the security advisory in #782) and landed both via #787, which is now merged. Your authorship is preserved on the cherry-picked commit (`f64f75e fix(http): preserve shared tenant sessions`). The behavior change ships in the upcoming **2.53.0** release with credit in CHANGELOG.

Closing this in favor of the merged work. Really appreciate the contribution — the shared-strategy bug was a real one for hosted multi-tenant deployments.

PR #784, opened 2026-05-12
fix(http): return 404 for stale sessions

LevSky22:
## Summary

Return HTTP `404` when a Streamable HTTP request includes a valid-looking `Mcp-Session-Id` that no longer maps to an active session.

The MCP Streamable HTTP transport spec says that after a server terminates a session, requests containing that old session ID must receive `404`, and clients must start a new session with a fresh `InitializeRequest`:

https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management

Returning `400` makes the stale-session case look like a malformed request instead of a recoverable expired session.

## Changes

- Return `404` for stale/expired Streamable HTTP session IDs.
- Keep `400` for missing session IDs on non-`initialize` requests.
- Keep notification handling unchanged: stale-session notifications still return `202` to avoid reconnect storms.
- Update session-management tests for stale request and mixed-batch request behavior.

## Validation

- `npx vitest run tests/unit/http-server-session-management.test.ts --coverage.enabled=false`
- `npm run build`


czlonkowski:
Thank you for the fix and the regression tests, @LevSky22! 🙏

We rebased this and #783 onto current `main` (your branch had drifted past several merges, including the security advisory in #782) and landed both via #787, which is now merged. Your authorship is preserved on the cherry-picked commit (`c642ea4 fix(http): return 404 for stale sessions`). The behavior change ships in the upcoming **2.53.0** release with credit in CHANGELOG.

Closing this in favor of the merged work. Really appreciate the contribution — spec-compliant `404` on terminated sessions lets MCP clients auto-recover instead of surfacing a "bad request" to the user.

Issue #777, opened 2026-05-05
get_workflow_details output exceeds MCP token limit — please add anthropic/maxResultSizeChars annotation


dendmyt:
## Problem

`get_workflow_details` returns duplicated node data (`workflow.nodes` + `activeVersion.nodes`), which causes large workflows (50k+ chars) to exceed Claude's MCP inline output limit. The result gets saved to a temp file on macOS (`/var/folders/...`) that is inaccessible from Claude's Linux bash sandbox — making the tool effectively unusable for complex workflows.

This affects ~50% of workflows in my n8n instance. Tested in both fresh and loaded sessions — same result.

## Fix

Add `_meta["anthropic/maxResultSizeChars"]` annotation (up to 500,000 chars) to the `get_workflow_details` tool definition, as documented in [Anthropic's MCP spec](https://code.claude.com/docs/en/whats-new/2026-w14).

Alternatively, deduplicating the node data (`workflow.nodes` and `activeVersion.nodes` appear to be identical) would reduce response size significantly and fix the issue without requiring any annotation.

## Environment

- Claude Cowork (desktop app, macOS)
- - n8n self-hosted on Railway
- - - Affected: any workflow with 50k+ chars in the response

czlonkowski:
Fixed in #786, released in [v2.53.0](https://github.com/czlonkowski/n8n-mcp/releases/tag/v2.53.0).

The duplicated payload came from n8n's draft/publish split (`workflow.nodes` is the editor draft, `workflow.activeVersion.nodes` is the published graph that's actually running — they diverge whenever there are unpublished edits). The fix:

- Strips `activeVersion` from `mode='full'` and `mode='details'` (keeps `activeVersionId` as a lightweight pointer) — roughly halves the response on active workflows.
- Adds `mode='active'` so the published/running graph is still reachable as a single-shaped response.
- Adds `_meta["anthropic/maxResultSizeChars"]: 450000` on `n8n_get_workflow` as defense in depth for genuinely huge workflows.

Auto-close didn't fire because the PR body wrote `Fixes **#777**` (bold markdown around the number), which GitHub's keyword parser doesn't recognize — closing manually now. Thanks for the detailed report and the link to the Anthropic spec.

PR #786, opened 2026-05-14
fix: strip duplicate activeVersion + add mode='active' for n8n_get_workflow (#777)

czlonkowski:
## Summary

Fixes **#777** — `n8n_get_workflow` responses on active workflows were exceeding Claude Code's default 25k-token MCP cap, causing the host to persist results to a temp file the model couldn't read. The duplication came from n8n's draft/publish model, which surfaces the published version under a nested `activeVersion` field alongside the editor draft.

This PR:

- **Strips the heavy `activeVersion` payload** from `mode='full'` and `mode='details'` responses while preserving the lightweight `activeVersionId` pointer. Cuts response size roughly in half on active workflows.
- **Adds `mode='active'`** to expose the published/running graph as a single-shaped response. Includes `activeVersionId`, `versionCreatedAt` (the version row's creation timestamp — within ~1s of the publish event in current n8n), and `versionName` at the top level. On older n8n versions that don't have the draft/publish split, it falls back to `workflow.nodes` when `active: true` so the mode stays usable across n8n versions. Returns `code: 'NO_ACTIVE_VERSION'` only for inactive workflows that were never published.
- **Adds `_meta["anthropic/maxResultSizeChars"]: 450000`** to the `n8n_get_workflow` tool definition per the [Claude Code MCP spec](https://code.claude.com/docs/en/mcp#raise-the-limit-for-a-specific-tool) — defense in depth for genuinely huge workflows even after dedup. Value is below the protocol's 500k-char ceiling to leave headroom for the MCP/JSON-RPC envelope.
- **Fixes a latent bug in `UIAppRegistry.injectToolMeta`** that overwrote rather than merged `_meta`, which would have silently dropped the size annotation on any future UI-app-enabled tool.

Reported by **@nepalez** — credit and thanks.

## Why a new mode instead of just stripping?

`activeVersion` is **not** a true duplicate. n8n now has a draft/publish split: `workflow.nodes` is the editor's working copy; `workflow.activeVersion.nodes` is what's actually running. They diverge whenever there are unpublished edits ("saving is not publishing", per n8n's own docs). Blindly stripping `activeVersion` would hide what's running in production. Splitting into separate modes preserves the semantic distinction:

- `mode='full'` / `'details'` → the draft (what you're editing)
- `mode='active'` → the published graph (what's actually running)

Verified live against staging workflow `9q68HobQ09ZaEjtw`, which has a draft change pending: the `Edit Fields` Set node's assignment value is `"this is unpublished"` in `mode='full'` and `""` (the published state) in `mode='active'`. Same node ID, different value — exactly the divergence the new mode is designed to surface.

## Test plan

- [x] `npm run typecheck` — clean
- [x] `npm run build` — clean
- [x] `npm test` (unit) — **4795 passing**, 35 skipped, 80.64% coverage
- [x] Unit tests for the strip behavior on `full` + `details` (including a regression guard that proves the draft, not the published version, is returned)
- [x] Unit tests for `handleGetWorkflowActive` covering: success path, fallback to workflow.nodes for older n8n / active orphans, `NO_ACTIVE_VERSION` for inactive + never-published, Zod / N8nApiError / generic error catches, and the defaults-to-null behavior for `versionCreatedAt`/`versionName`
- [x] Unit test asserting `mode='active'` routes to `handleGetWorkflowActive` in the server dispatch
- [x] Unit test asserting the tool def carries the `mode='active'` enum value and the `_meta` size annotation
- [x] Unit test asserting `UIAppRegistry.injectToolMeta` preserves pre-existing `_meta` keys
- [x] Integration tests at `tests/integration/n8n-api/workflows/get-workflow{,-details}.test.ts` updated to assert `activeVersion` is absent
- [x] Manual smoke test against live staging (`https://n8n-test.n8n-mcp.com/workflow/9q68HobQ09ZaEjtw`):
  - `mode='full'` → `activeVersion` stripped, `activeVersionId` retained, draft value visible
  - `mode='details'` → same shape nested under `data.workflow`, executionStats present
  - `mode='active'` → published graph as top-level nodes, `versionCreatedAt` + `versionName` present, draft assignment not visible
  - `mode='active'` on never-activated workflow → returns `NO_ACTIVE_VERSION` with a helpful message

## Files changed

- `src/types/n8n-api.ts` — `ActiveWorkflowVersion` interface; typed `activeVersion` / `activeVersionId` on `Workflow`
- `src/types/index.ts` — broadened `ToolDefinition._meta` to permit `anthropic/maxResultSizeChars` and other keys
- `src/mcp/handlers-n8n-manager.ts` — `stripActiveVersion` helper; applied in `handleGetWorkflow` and `handleGetWorkflowDetails`; new `handleGetWorkflowActive` with older-n8n fallback
- `src/mcp/server.ts` — wired `mode='active'` into the `n8n_get_workflow` switch
- `src/mcp/tools-n8n-manager.ts` — added `'active'` to the mode enum, updated description, added `_meta` annotation
- `src/mcp/ui/registry.ts` — `injectToolMeta` now spread-merges
- `src/mcp/tool-docs/workflow_management/n8n-get-workflow.ts` — documented `active` mode + draft/publish distinction
- `src/mcp/tools-documentation.ts` — one-liner updated
- `tests/unit/mcp/handlers-n8n-manager.test.ts`, `tests/unit/mcp/get-workflow-tool.test.ts` (new), `tests/unit/mcp/get-workflow-mode-dispatch.test.ts` (new), `tests/unit/mcp/ui/registry.test.ts`
- `tests/integration/n8n-api/workflows/get-workflow{,-details}.test.ts`
- `CHANGELOG.md`, `package.json`, `package-lock.json` — v2.53.0

## Out of scope

Other potentially large-output tools (`get_template`, `search_nodes` with full options, `n8n_list_workflows`, `get_node_info`) could benefit from similar `_meta` annotations. Tracking that as a follow-up rather than expanding this PR.


czlonkowski:
Follow-up commit `20fa5cf` addresses the Copilot review comments + code-reviewer/code-simplifier passes.

### Acted on

- **Copilot #2 (`publishedAt` semantics):** renamed to `versionCreatedAt` and documented honestly — the value is `activeVersion.createdAt` (the version row's creation time), not the publish-event time (`workflowPublishHistory[?event=activated].createdAt`, which is ~1s later in current n8n). Defaults to `null` when source is missing.
- **Code-simplifier:** `stripActiveVersion` is no longer generic and dropped the dead falsy guard (callers always pass a real `Workflow`). `ActiveWorkflowVersion` pruned to only the fields the handler actually reads (`nodes`, `connections`, `name`, `createdAt`), with an open `[key: string]: unknown` index so extra wire keys don't break the type.
- **Code-reviewer (size cap):** `maxResultSizeChars` lowered from 500_000 (the protocol ceiling) to 450_000 to leave ~10% headroom for the MCP/JSON-RPC envelope.
- **Code-reviewer (test gap):** added two regression tests in `tests/unit/mcp/handlers-n8n-manager.test.ts`:
  - `mode='full'` must return the **draft** `nodes`/`connections` after the strip, not the published version's (would catch someone wiring `stripActiveVersion` to also overwrite `nodes`).
  - Pre-draft/publish n8n versions (no `activeVersion` key at all) must pass through unmangled.

### Declined (with reasons)

- **Copilot #1 (dist for ui/registry not updated):** `dist/mcp/ui/*` has **never** been tracked in git (confirmed via `git ls-files dist/mcp/ui/`). It's a pre-existing gap in the repo's tracked dist set, not introduced by this PR. Runtime is unaffected: Docker/CI rebuild dist from source, and npm publish reruns `npm run build`. Out of scope.
- **Copilot #3 (fallback for older n8n without draft/publish):** `mode='active'` is a brand-new mode in this PR — nobody had it before. Pre-draft/publish n8n users still get the default `mode='full'` behavior they always had. Adding a fallback would muddy the semantic ("active means published" vs "active means whatever is running").
- **Reviewer (guard against malformed `activeVersion.nodes/connections`):** n8n's `WorkflowHistory` entity declares these as required columns. Project convention is "don't validate scenarios that can't happen" (CLAUDE.md).
- Other nits (distinguishing orphan from never-activated, soft-breaking changelog note) deferred — low value relative to PR scope.

### Verification

- `npm run typecheck` ✓
- `npm test` ✓ (4786 passing, +2 new regression tests)
- `npm run build` ✓
- Manual smoke on staging unchanged — same draft/active divergence on `9q68HobQ09ZaEjtw`.

czlonkowski:
Follow-up commit `f1a852d` addresses the Copilot pass on `e956210`.

### Acted on

- **CHANGELOG mismatch:** corrected the documented size cap to `450000` (was still showing the pre-review `500000`).
- **PR description vs code:** updated to reference `versionCreatedAt` everywhere (was lingering `publishedAt`).
- **Older n8n fallback** (`handlers-n8n-manager.ts:838`): `handleGetWorkflowActive` now falls back to `workflow.nodes` / `workflow.connections` when `workflow.active === true` but `activeVersionId`/`activeVersion` are missing. Pre-draft/publish n8n versions return no `activeVersion` field at all and the workflow body itself is the running graph — refusing those requests with `NO_ACTIVE_VERSION` made the mode unusable on those instances. The same branch also covers the rare orphan case in newer n8n where `activeVersionId` got nulled but `active` is still true. `NO_ACTIVE_VERSION` now means "inactive **and** never published" specifically. Tests updated to pin both the fallback and the genuine never-activated case; tool doc updated.
- **`dist/mcp/ui/registry.*` not tracked:** force-added the four files so the gap is closed. The release pipeline regenerates them via `npm run build:all` before `npm publish`, so the published npm artifact already had the fix — but tracking them avoids the false-positive on future PRs.

### Verification

- `npm run typecheck` ✓
- `npm test` ✓ (4795 passing — net +2 after replacing 2 obsolete tests with 4 fallback tests)
- `npm run build` ✓

Issue #654, opened 2026-03-22
handleRequest returns 400 for notifications with stale sessions, causing Claude disconnect storms


czlonkowski:
## Problem

`handleRequest()` in `http-server-single-session.js` treats JSON-RPC **notifications** (e.g. `notifications/initialized`, `notifications/cancelled`) identically to tool calls when the session ID is unknown — it returns HTTP 400 with `"Session not found or expired"`. This causes Claude's proxy to interpret the connection as broken, showing users a "disconnected" status and triggering reconnection storms.

**Impact:** 930 errors/day at peak, 216 affected users. Ongoing since ~v2.37.3.

## Root Cause

Claude's proxy (claudemcpcontent.com) maintains two parallel MCP connections. When one connection creates a new session, the other may still send `notifications/initialized` with the old session ID. Since `handleRequest()` doesn't distinguish notifications from requests, it returns 400:

```
Timeline:
1. Connection A: initialize → new session "sess-123" created
2. Connection A: notifications/initialized (sess-123) → SUCCESS  
3. Connection B: notifications/initialized (sess-OLD) → 400 "Session not found"
   → Claude shows "disconnected"
   → User/Claude reconnects → new session created
   → Repeat (reconnection storm)
```

The relevant code path (`http-server-single-session.js` lines 391-414):

```javascript
else {
    // Falls here when sessionId exists but this.transports[sessionId] is undefined
    let errorMessage = 'Bad Request: No valid session ID provided and not an initialize request';
    if (sessionId && !this.isValidSessionId(sessionId)) {
        errorMessage = 'Bad Request: Invalid session ID format';
    } else if (sessionId && !this.transports[sessionId]) {
        errorMessage = 'Bad Request: Session not found or expired';
    }
    res.status(400).json({
        jsonrpc: '2.0',
        error: { code: -32000, message: errorMessage },
        id: req.body?.id || null
    });
}
```

No distinction is made between notifications (fire-and-forget, no `id` field) and requests (expecting a response).

## Expected Behavior

Per JSON-RPC 2.0 spec, **notifications do not expect a response**. The MCP SDK itself already handles this correctly in `StreamableHTTPServerTransport.handlePostRequest()` — when a POST contains only notifications/responses, it returns `202 Accepted` (lines 368-375 of `streamableHttp.js`). However, the session validation in `handleRequest()` runs **before** the message type is checked, so notifications with stale sessions never reach this code.

[...]

PR #787, opened 2026-05-14
fix(http): session handling — 404 for stale, preserve shared tenant (rebased #783 + #784)

czlonkowski:
## Summary

Rebases two community fixes from @LevSky22 ([#783](https://github.com/czlonkowski/n8n-mcp/pull/783), [#784](https://github.com/czlonkowski/n8n-mcp/pull/784)) onto current `main`. Their original branches were forked from an older `main` and their raw diffs would have reverted #782 (security telemetry redaction, GHSA-f3rg-xqjj-cj9w), #785 (n8n 2.20.x bump + community-node `typeVersion` utility), and #786 (`n8n_get_workflow` `activeVersion` fix), so we cherry-pick the two surgical commits instead. Authorship preserved on both fix commits via `cherry-pick -x`.

### `fix(http): preserve shared tenant sessions` — #783

The eager same-instance session cleanup at the top of the HTTP `initialize` handler ran *before* the configured `MULTI_TENANT_SESSION_STRATEGY` was consulted. When the strategy was `shared` — intended to let multiple MCP clients reuse the same tenant context concurrently — a second client's `initialize` for the same `x-instance-id` still wiped the first client's session, leaving it to fail subsequent requests with `Session not found or expired`. Cleanup is now gated behind `ENABLE_MULTI_TENANT=true && MULTI_TENANT_SESSION_STRATEGY=instance`, restoring the documented behavior of `shared`. The `instance` strategy is unchanged.

### `fix(http): return 404 for stale sessions` — #784

The [MCP Streamable HTTP spec (2025-06-18)](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management) requires servers to respond with `404` (not `400`) when a request carries a valid-format `Mcp-Session-Id` that no longer maps to an active session — clients use that as the signal to start a new session. The two affected sites (TOCTOU window after the session-existence check, and the regular non-`initialize` fall-through) now return `404` with `"Session not found or expired"`. `400` is preserved for genuinely bad requests (missing session ID on non-`initialize`, malformed format) and `202` is preserved for stale-session notifications (the anti-reconnect-storm path from #654).

The two fixes compose well: in `instance` mode, when the eager cleanup terminates an old session, the client now receives a spec-compliant `404` and can transparently re-initialize.

## Validation

- `npm run typecheck` clean
- `npx vitest run tests/unit/http-server-session-management.test.ts` — 51/51 (4 new tests from the cherry-picks, shared-mode assertion subsequently tightened to mirror the instance-mode test)
- `npm test` — only two pre-existing flaky failures unrelated to this branch (timing-flaky DB perf test and an integration test against the local n8n instance; both pass on `main` when re-run)
- `n8n-mcp-tester` agent smoke pass against the rebuilt stdio MCP server — green
- Direct HTTP curl verification against a local HTTP server in both `shared` and `instance` strategies, covering all status-code paths from both fixes — all expected codes returned

## Notes

- Version is already at 2.53.0 (queued release); CHANGELOG entries added under the existing 2.53.0 `### Fixed` section.
- `code-reviewer` raised five non-blocking observations; the actionable one (loose `not.toHaveBeenCalledWith(400)` assertion in the new shared-mode test) is fixed in `test(http-session): tighten shared-mode assertion`. The remaining suggestions are documentation/comment nits for follow-up.
- After this lands, the original PRs #783 and #784 will be closed with thank-you comments linking here.


czlonkowski:
Thanks Copilot. Addressing all three:

1. **CHANGELOG.md:15 — "malformed format" is unreachable** — agreed, fixed in b38bc82. `isValidSessionId()` accepts any non-empty string, so the changelog now reads "400 for genuinely bad requests (missing session ID on non-initialize)" and drops the dead format-validation phrasing.

2. **src/http-server-single-session.ts:744 — TOCTOU guard appears unreachable** — technically correct as the code stands today (no `await` between the `this.transports[sessionId]` membership check at L698 and the local assignment at L730, so V8 cannot interleave a cleanup). But the guard is pre-existing structure that LevSky22's cherry-pick did not introduce — the only change at that site is the 400→404 status code update. The surrounding handler runs inside `consoleManager.wrapOperation` and the broader transport plumbing does take awaits elsewhere, so we'd rather preserve the defensive read-and-recheck pattern in case a future edit introduces an async boundary above it. Leaving as-is for this PR; happy to revisit as a separate cleanup that simplifies the whole `else if` branch.

3. **src/http-server-single-session.ts:789 — log/comment text mentions "malformed"** — same underlying observation as (1) (`isValidSessionId` is permissive, so the "Invalid session ID format" branch at L785-786 is effectively dead). The comment/log wording at L771-781 is from the upstream cherry-pick (#784) and we'd prefer not to diverge from the contributor's text for a doc-level nit. The actionable user-facing copy (CHANGELOG) is fixed; in-source wording can be tightened in a follow-up alongside any decision about whether to implement real format validation in `isValidSessionId()`.

PR #794, opened 2026-05-19
fix(diff-engine): accept undefined as property-removal marker (closes #292)

### LINKED ISSUE #292: [Documentation] Missing: How to remove node properties using updateNode (undefined pattern)
Hej, wrzucam kolejne zgłoszenie po tym jak ai sobie nie radził i robiliśmy śledztwo. Najpierw pełne zgłoszenie napisane przez Claude po całej sprawie, potem podrzucę historię.
// Zbyszek

### 🏷️ Title:
```
[Documentation] Missing: How to remove node properties using updateNode (undefined pattern)
```

---

### 📝 Description:

```markdown
## Problem Description

The documentation for `n8n_update_partial_workflow` with `updateNode` operations doesn't explain how to **remove** properties from nodes. This leads to confusion when trying to migrate from deprecated properties (e.g., `continueOnFail` → `onError`).

## What We Tried

**Scenario**: Update HTTP Request node from deprecated `continueOnFail: true` to modern `onError: "continueRegularOutput"`

**Attempt 1** - Only set the new property:
```javascript
{
  type: "updateNode",
  updates: {
    onError: "continueRegularOutput"
  },
  nodeName: "getUnmanagedGroups"
}
```

**Expected**: Old `continueOnFail` would be removed or ignored  
**Actual**: Both properties remained on the node, causing validation error:
```
"Cannot use both continueOnFail and onError properties. Use only onError for modern workflows."
```

**Attempt 2** - Explicitly set deprecated property to false:
```javascript
{
  type: "updateNode",
  updates: {
    onError: "continueRegularOutput",
    continueOnFail: false
  },
  nodeName: "getUnmanagedGroups"
}
```

**Expected**: Setting to `false` would disable the old behavior  
**Actual**: Same validation error - validator checks for **presence** of both properties, not their values

## Our Investigation

We used DeepWiki to analyze the n8n-mcp source code and found:

### 1. The Solution Exists in Code
`src/services/workflow-auto-fixer.ts` (L311-317)
```typescript
const operation: UpdateNodeOperation = {
  type: 'updateNode',
  nodeId: nodeName,
  updates: {
    onError: undefined // This will remove the property
  }
};
```

**The pattern exists but is undocumented for users.**

### 2. Validation Logic
`src/services/workflow-validator.ts` (L1385-1393)
```typescript
if (node.continueOnFail !== undefined && node.onError !== undefined) {
    result.errors.push({
      message: 'Cannot use both "continueOnFail" and "onError" properties.'
    });
}
```

Validator checks for property **existence**, not values. Both `true` and `false` trigger the error.

### 3. Why This Isn't Intuitive
Most REST APIs use "update = replace" semantics. Users naturally assume:
- Not mentioning a property = remove it
- Setting to `false` = disable it

But n8n-mcp uses **selective merge** - only provided properties are touched, others remain unchanged.

### Add Documentation Section: "Removing Node Properties"

Location: `tools_documentation` for `n8n_update_partial_workflow`, after "Update node parameter" example

**Suggested content**:

```markdown
### Removing Properties

To remove a property from a node, explicitly set it to `undefined`:

**Example 1: Migrate from deprecated continueOnFail to modern onError**
```javascript
n8n_update_partial_workflow({
  id: "workflow-id",
  operations: [{
    type: "updateNode",
    nodeName: "HTTP Request",
    updates: {
      onError: "continueRegularOutput",
      continueOnFail: undefined  // Removes deprecated property
    }
  }]
})
```

**Example 2: Remove error handling completely**
```javascript
n8n_update_partial_workflow({
  id: "workflow-id",
  operations: [{
    type: "updateNode",
    nodeName: "MyNode",
    updates: {
      onError: undefined,
      retryOnFail: undefined,
      maxTries: undefined
    }
  }]
})
```

**Note**: The `updateNode` operation uses selective merge - it only updates properties you explicitly provide. Properties not mentioned remain unchanged. Use `undefined` to explicitly remove properties.
```

### Additional Context

**Why this matters**:
1. Common migration scenario (n8n deprecated `continueOnFail` in favor of `onError`)
2. Validator enforces mutual exclusivity but docs don't explain removal pattern
3. Pattern is used internally (auto-fixer) but not exposed to users
4. Natural assumptions (update = replace) lead to frustration

**Related documentation**:
- Changelog mentions error handling update (2025-07-09) but doesn't explain removal pattern
- Node-specific validator warns about deprecated properties but doesn't show how to remove them

## Environment
- n8n-mcp version: 2.18.1
- n8n instance: https://n8n.bitback.cloud
- Node.js: v22.15.0

## Workaround
Once we discovered the `undefined` pattern through source code analysis, it worked perfectly:

```javascript
{
  type: "updateNode",
  updates: {
    onError: "continueRegularOutput",
    continueOnFail: undefined  // ✅ This removes the property
  },
  nodeName: "getUnmanagedGroups"
}
```

The workflow validated successfully after this change.
```

---

**Gotowe do wklejenia.** Wszystko co trzeba w jednym issue - problem, śledztwo, rozwiązanie, sugestia poprawki.

justadityaraj:
## Summary

`setNestedProperty` in `WorkflowDiffEngine` only treats `null` as a deletion signal. But `workflow-auto-fixer.ts` already uses `{onError: undefined}` to remove properties — with the literal comment `// This will remove the property`. That comment isn't true today: the diff engine sets the property to `undefined` instead of deleting it, leaving `hasOwnProperty(key)` true.

This PR makes `undefined` behave identically to `null` for property removal, so both the user-facing `n8n_update_partial_workflow` API and the internal auto-fixer actually do what they say.

## What changes

- `src/services/workflow-diff-engine.ts` — `setNestedProperty` now treats `value === null || value === undefined` as the deletion marker, at both the intermediate-path guard and the final-key write site
- `tests/unit/services/workflow-diff-engine.test.ts` — adds a new `describe('undefined value property deletion (Issue #292)')` block with 5 tests mirroring the existing null-deletion block, including the real `continueOnFail → onError` migration scenario from #292
- `src/mcp/tool-docs/workflow_management/n8n-update-partial-workflow.ts` — one-line addition to the existing "Removing Properties with null" section noting that `undefined` is also accepted

## Why this is the right fix

Three signals lined up here:

1. **Internal inconsistency** — `workflow-auto-fixer.ts` (around the `processErrorOutputFixes` path) emits `{onError: undefined}` with a comment promising removal. Before this PR, that promise was silently broken
2. **User confusion in #292** — the original reporter and a second user both reached for `undefined` based on natural JS semantics; docs eventually pinned the answer to `null`, but the underlying mismatch with the auto-fixer remained
3. **No behavioral regression** — `null` still deletes; the type signature (`[path: string]: any`) already permits `undefined`; existing 215 diff-engine tests stay green

## Tests

- Mirror tests for `undefined` covering: simple removal, nested removal (`'parameters.authentication': undefined`), no-op on missing property, no-op on missing parent path, full `continueOnFail → onError` migration
- All 220 tests in `workflow-diff-engine.test.ts` pass (215 existing + 5 new)
- All 18 tests in `workflow-auto-fixer.test.ts` pass
- 2118/2118 tests across `tests/unit/services/` pass
- `npm run build` clean

## Closes

#292

## Test plan
- [x] New tests pass: `npx vitest run tests/unit/services/workflow-diff-engine.test.ts -t "undefined value property deletion"`
- [x] Diff engine + auto-fixer suites pass: `npx vitest run tests/unit/services/workflow-diff-engine.test.ts tests/unit/services/workflow-auto-fixer.test.ts`
- [x] Full services unit tests pass: `npx vitest run tests/unit/services/`
- [x] Build clean: `npm run build`

### DIFF
--- a/src/mcp/tool-docs/workflow_management/n8n-update-partial-workflow.ts
+++ b/src/mcp/tool-docs/workflow_management/n8n-update-partial-workflow.ts
@@ -201,7 +201,7 @@ Please choose a different name.
 
 ## Removing Properties with null
 
-To remove a property from a node, set its value to \`null\` in the updates object. This is essential when migrating from deprecated properties or cleaning up optional configuration fields. \`undefined\` is also accepted and behaves identically — both delete the property from the node.
+To remove a property from a node, set its value to \`null\` in the updates object. This is essential when migrating from deprecated properties or cleaning up optional configuration fields. Over the MCP/JSON-RPC API always use \`null\` — an \`undefined\` value is dropped by JSON serialization before it reaches the server, so it would be a silent no-op. (Internally, in-process callers such as the workflow auto-fixer may pass \`undefined\`, which the diff engine now treats the same as \`null\`.)
 
 ### Why Use null?
 - **Property removal**: Setting a property to \`null\` removes it completely from the node object
--- a/src/mcp/tool-docs/workflow_management/n8n-update-partial-workflow.ts
+++ b/src/mcp/tool-docs/workflow_management/n8n-update-partial-workflow.ts
@@ -201,7 +201,7 @@ Please choose a different name.
 
 ## Removing Properties with null
 
-To remove a property from a node, set its value to \`null\` in the updates object. This is essential when migrating from deprecated properties or cleaning up optional configuration fields.
+To remove a property from a node, set its value to \`null\` in the updates object. This is essential when migrating from deprecated properties or cleaning up optional configuration fields. \`undefined\` is also accepted and behaves identically — both delete the property from the node.
 
 ### Why Use null?
 - **Property removal**: Setting a property to \`null\` removes it completely from the node object
--- a/src/services/workflow-diff-engine.ts
+++ b/src/services/workflow-diff-engine.ts
@@ -1665,7 +1665,8 @@ export class WorkflowDiffEngine {
       if (!Object.prototype.hasOwnProperty.call(current, key)
           || typeof current[key] !== 'object'
           || current[key] === null) {
-        if (value === null) return; // parent path doesn't exist, nothing to delete
+        // null or undefined signals deletion — if parent path doesn't exist there's nothing to delete
+        if (value === null || value === undefined) return;
         current[key] = {};
       }
       current = current[key];
@@ -1676,7 +1677,11 @@ export class WorkflowDiffEngine {
     if (DANGEROUS_PATH_KEYS.has(finalKey)) {
       throw new Error(`Invalid property path: "${path}" contains a forbidden key`);
     }
-    if (value === null) {
+    // Both null and undefined remove the property. undefined is accepted because
+    // workflow-auto-fixer.ts already uses `{prop: undefined}` to signal removal
+    // (see processErrorOutputFixes), so treating only null as the deletion marker
+    // left those fixes silently inert at the diff-engine layer.
+    if (value === null || value === undefined) {
       delete current[finalKey];
     } else {
       current[finalKey] = value;
--- a/tests/unit/services/workflow-diff-engine.test.ts
+++ b/tests/unit/services/workflow-diff-engine.test.ts
@@ -6158,6 +6158,127 @@ describe('WorkflowDiffEngine', () => {
     });
   });
 
+  describe('undefined value property deletion (Issue #292)', () => {
+    // Mirrors the `null value property deletion` block. `undefined` is accepted
+    // as a deletion marker because workflow-auto-fixer.ts already uses
+    // `{prop: undefined}` to remove properties (see processErrorOutputFixes) —
+    // before this, those fixes set the property to `undefined` instead of
+    // deleting it, which left `hasOwnProperty(prop)` true and could trip
+    // validators that check property presence rather than value.
+
+    it('should delete a property when value is undefined', async () => {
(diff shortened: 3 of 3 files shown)


justadityaraj:
Friendly bump on this — happy to address any review feedback or rebase if helpful. The fix is narrow (engine accepts `undefined` the same as `null` for property removal, matching what `workflow-auto-fixer.ts` already passes in) with 5 mirror tests added.

czlonkowski:
Thanks @justadityaraj — nice catch on a real latent bug. The auto-fixer pushes `{onError: undefined}` (with a comment claiming it removes the property), but the diff engine was setting the property to `undefined` rather than deleting it, leaving `'onError' in node` true. Treating `undefined` like `null` is the right fix, and the 5 mirror tests are solid.

One adjustment pushed to the branch (and rebased onto current `main`): the user-facing doc line said `undefined` "behaves identically" to `null` over the API. Over MCP/JSON-RPC that's misleading — `JSON.stringify` drops `undefined`-valued keys before they reach the server, so it'd be a silent no-op for external callers. I reworded it to tell API users to use `null`, and noted that in-process callers (the auto-fixer) may pass `undefined`, which the engine now treats the same as `null`.

`npm run typecheck` and the full `workflow-diff-engine` suite (220 passing) are green. Recommending merge once CI confirms.

PR #795, opened 2026-05-21
fix: scope Code node primitive return checks

AjTheSpidey:
﻿This tightens the Code node return validator a bit.

The previous fix for helper functions avoided false positives by skipping primitive-return checks whenever a helper function was present. That fixed the noisy case, but it also meant a snippet like this could slip through:

```js
const isValid = (item) => { return false; };
if (!items.length) return "empty";
```

This keeps ignoring returns inside nested helper function bodies, but still checks returns from the actual Code node flow, including returns inside top-level try/catch blocks.

I ran:

```
node node_modules\vitest\vitest.mjs run tests\unit\services\node-specific-validators.test.ts --coverage.enabled=false
node node_modules\typescript\bin\tsc --noEmit --pretty false
node node_modules\typescript\bin\tsc -p tsconfig.build.json --pretty false
```

Small note: `npm ci` currently fails for me on main because the lockfile looks out of sync with `package.json`, so I installed without rewriting the lockfile to run the checks.

### DIFF
--- a/src/services/node-specific-validators.ts
+++ b/src/services/node-specific-validators.ts
@@ -27,6 +27,21 @@ const MAX_CODE_LENGTH = 200_000;
  */
 const MAX_SHORT_INPUT_LENGTH = 2_000;
 
+/**
+ * Max distance the function-head detector scans backward from a `{` to find the
+ * matching `(`. A real parameter list is short; capping the walk keeps brace
+ * scanning linear on adversarial input (e.g. many unmatched `) {`).
+ */
+const MAX_PARAM_SCAN = 2_000;
+
+/**
+ * Detects a top-level primitive return in a JS Code node. Keyword literals
+ * require a trailing word boundary so identifiers that merely start with one
+ * (e.g. `return trueItems`) are not misflagged. Module-level + flagless so it
+ * can be reused without `lastIndex` state.
+ */
+const JS_PRIMITIVE_RETURN_RE = /return\s+(?:(?:true|false|null|undefined)\b|\d+|['"`])/m;
+
 export interface NodeValidationContext {
   config: Record<string, any>;
   errors: ValidationError[];
@@ -1389,7 +1404,8 @@ export class NodeSpecificValidators {
     // Detect a *real* top-level return. For JS, scan the stripped view so a
     // return that only appears inside a comment, string, or nested function
     // body (e.g. `// return "x"`) does not satisfy the "must return data" check.
-    const returnScanCode = language === 'javaScript'
+    // Skip the strip for very large code (mirrors hasTopLevelPrimitiveReturn).
+    const returnScanCode = (language === 'javaScript' && code.length <= MAX_CODE_LENGTH)
       ? this.stripNestedJavaScriptFunctionBodies(code)
       : code;
     const hasReturn = /return\s+/.test(returnScanCode);
@@ -1479,11 +1495,11 @@ export class NodeSpecificValidators {
 
   private static hasTopLevelPrimitiveReturn(code: string): boolean {
     if (code.length > MAX_CODE_LENGTH) {
-      return /return\s+(true|false|null|undefined|\d+|['"`])/m.test(code);
+      return JS_PRIMITIVE_RETURN_RE.test(code);
     }
 
     const topLevelCode = this.stripNestedJavaScriptFunctionBodies(code);
-    return /return\s+(true|false|null|undefined|\d+|['"`])/m.test(topLevelCode);
+    return JS_PRIMITIVE_RETURN_RE.test(topLevelCode);
   }
 
   private static stripNestedJavaScriptFunctionBodies(code: string): string {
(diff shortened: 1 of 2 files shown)


czlonkowski:
Thanks @AjTheSpidey — replacing the coarse "skip the check whenever any helper exists" heuristic with a real brace-depth scanner is the right direction, and the string/comment/template handling is solid.

While reviewing I confirmed (by running the validator on real snippets) that the scanner still produced **false positives** on three common, valid Code-node patterns — the exact failure class this PR targets — because their bodies weren't recognized as function bodies and the inner `return <primitive>` leaked into the top-level scan:

- object-method shorthand and class methods — `const o = { foo() { return 1; } }`
- generator functions — `function* g() { return 1; }`
- regex literals containing braces — `s.replace(/}/g, '')` (the `}` was counted as a real brace, desyncing depth)

I pushed fixes directly to the branch (and rebased onto current `main`):

- `startsJavaScriptFunctionBody` now also recognizes method shorthand / class methods / getter-setter / generator heads, while explicitly excluding control-flow keywords (`if`/`for`/`while`/`switch`/`catch`/`with`) so `if (x) {` isn't mistaken for a function body.
- `stripNestedJavaScriptFunctionBodies` gained a `regex` scanner state with `[...]` char-class handling and a **newline bail-out** that bounds any division-vs-regex misclassification to a single line.

Added 5 regression tests (the three cases above + a regex-literal-with-char-class case + a guard that a *real* primitive return next to a regex literal is still flagged). Full `node-specific-validators` suite is green (197 passing) and `npm run typecheck` is clean. Recommending merge once CI confirms.

czlonkowski:
Follow-up from a Codex review pass: the scanner had one remaining false positive I've now fixed on the branch.

`startsJavaScriptFunctionBody` matched the parameter list with a greedy `\([^)]*\)`, which stops at the first `)`. So a helper whose params contain nested parens — e.g. a default value that calls a function:

```js
function normalize(item = $input.first()) { return null; }
return [{json: {v: normalize()}}];
```

wasn't recognized as a function body, and its inner `return null` leaked into the top-level scan → false `Cannot return primitive values directly`. (This same `\([^)]*\)` limitation was in the original regex too.)

Fix: locate the parameter list's matching `(` by scanning backward with a paren counter, then test the preceding head for a function/method signature. Added regression tests for named-function, arrow, and method helpers with function-call default params, plus a guard that a real top-level primitive return alongside such a helper is still flagged. Full `node-specific-validators` suite green (199 passing) and `npm run typecheck` clean.

czlonkowski:
A second Codex pass on the updated branch found two more false negatives — both now fixed:

1. **`for await` misdetected as a function body.** `for await (const x of y) { return "bad"; }` — the head's trailing word is `await`, which my method-shorthand detection captured, so the loop body got stripped and the primitive top-level return was missed. Added `await` to the control-flow exclusion set.
2. **Commented/strung-out returns satisfied the "must return data" check.** The presence check used the raw code, so a `// return "x"` with no real return slipped past. The JS return-presence check now uses the stripped top-level view (consistent with the primitive-return check).

Added regression tests for both (plus a valid for-await case). Full `node-specific-validators` suite green (202 tests), and the broader `tests/unit/services` + `tests/unit/mcp` suites (74 files) pass with no regressions.

czlonkowski:
A final Codex pass caught two more issues — both now fixed:

1. **Quadratic scan (DoS risk).** The backward search for a function's matching `(` could walk to the start of the file for every `{` on adversarial input (many unmatched `) {`), and it runs per-brace — O(n²), bypassing the `MAX_CODE_LENGTH` guard. Capped the walk at `MAX_PARAM_SCAN` (2000) chars so brace scanning stays linear, and skipped the strip on the missing-return path for oversized code. An adversarial `) {`×40000 input now validates in ~90ms instead of hanging.
2. **`return trueItems` false positive.** The primitive-return regex had no word boundary, so an identifier starting with a primitive keyword matched (`return true` inside `return trueItems`). Hoisted the pattern to one module constant with a `\b` after the keyword literals; `return trueItems` / `falseyVar` / `nullableData` / `undefinedish` are no longer flagged, while real `return true/null/42/"x"` still are.

Added a regression test for the identifier case. Full `node-specific-validators` suite green (203 tests) and `tests/unit/services` + `tests/unit/mcp` (74 files) pass with no regressions.

This was the third Codex pass on this PR; no further findings expected, but happy to re-run.

czlonkowski:
A 4th Codex pass found two more narrow edge cases — both fixed:

1. **Division after a string/template literal** (`const ratio = "10" / 2;`) was misread as a regex-literal start, which blanked the rest of the line and dropped the real top-level return → false "must return data". A closing quote is now treated as a value, so the `/` is division.
2. **A block comment between `)` and `{`** in a helper (`function f() /* note */ {`) defeated function-head detection → the helper's primitive return leaked to the top level → false "Cannot return primitive". The backward scan now skips block comments too.

Regression tests added for both. Full `node-specific-validators` suite green (205) + `tests/unit/services`/`tests/unit/mcp` (74 files) clean.

Note on scope: the Code-node scanner is a heuristic, and each review pass has surfaced progressively rarer edges (methods/generators/regex → for-await/comment-only → quadratic/word-boundary → division-after-string/comment-before-brace). All are now handled, and every fix has a test. The PR is a clear net improvement over the previous "skip the check whenever any helper exists" behavior. One known residual: a `//` line comment placed *directly* between `)` and `{` isn't skipped (detecting that backward can't be done safely without full tokenization, and it risks misfiring on URL strings in default params) — extremely rare and low-impact (worst case a single false primitive-return hint on contrived code).

PR #813, opened 2026-06-02
fix: strip nodeGroups + sourceWorkflowId in cleanWorkflowForUpdate (n8n 2.x PUT rejects them)

DUC750:
## Problem

`cleanWorkflowForUpdate()` does not strip two read-only root fields that recent n8n versions return on `GET /workflows` but **reject** on `PUT /workflows` (the workflow schema is `additionalProperties: false`):

- **`nodeGroups`** — returned by n8n 2.x
- **`sourceWorkflowId`** — added in n8n 2.25+ (migration `AddSourceWorkflowId`)

Because the MCP server reconstructs the full workflow and PUTs it back, every workflow update fails with:

```
400 request/body must NOT have additional properties
```

This breaks all `n8n_update_*_workflow` operations against current n8n (reproduced on **n8n 2.57.1**).

## Verification

Enumerated empirically which root fields n8n 2.57.1's `PUT /workflows` rejects by sending the GET body field-by-field:

```
rejected: updatedAt, createdAt, id, description, active, isArchived, meta,
          nodeGroups, versionId, activeVersionId, versionCounter, triggerCount,
          sourceWorkflowId, shared, tags, activeVersion
accepted: name, nodes, connections, settings, staticData, pinData
```

`sourceWorkflowId` was the only rejected field not already stripped by the destructure.

## Change

- Add `nodeGroups` and `sourceWorkflowId` to the `cleanWorkflowForUpdate()` strip destructure.
- Extend the "remove all read-only and computed fields" unit test to assert both (plus `activeVersionId` / `activeVersion`).

All 104 tests in `tests/unit/services/n8n-validation.test.ts` pass.

### DIFF
--- a/src/services/n8n-validation.ts
+++ b/src/services/n8n-validation.ts
@@ -176,6 +176,10 @@ export function cleanWorkflowForUpdate(workflow: Workflow): Partial<Workflow> {
     // (workflow schema has additionalProperties:false) -> must be stripped,
     // otherwise every update fails with 400 "must NOT have additional properties".
     nodeGroups,
+    // n8n 2.25+ (migration AddSourceWorkflowId) returns `sourceWorkflowId` on GET
+    // but the PUT /workflows schema (additionalProperties:false) rejects it too.
+    // Without stripping, every workflow update fails with 400 on n8n >= 2.25.
+    sourceWorkflowId,
     // Keep everything else
     ...cleanedWorkflow
   } = workflow as any;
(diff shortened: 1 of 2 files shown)


czlonkowski:
Closing as superseded by #840, which shipped in **v2.57.2**. We inverted `cleanWorkflowForUpdate()` to an allowlist (`name`, `nodes`, `connections`, `settings` only), so both fields your PR targeted — `nodeGroups` and `sourceWorkflowId` — are stripped, along with the n8n 2.x versioning fields (`activeVersion`, `activeVersionId`, `versionCounter`) and anything n8n adds in the future.

Thanks for the fix and for spotting `sourceWorkflowId` as well — that confirmed the denylist approach couldn't keep up and pushed us toward the allowlist. Appreciated!

PR #830, opened 2026-06-03
fix(credentials): paginate past 100 credentials in n8n_manage_credentials (#816)

### LINKED ISSUE #816: n8n_manage_credentials: cannot paginate past 100 credentials (no cursor param); breaks get-by-id and includeUsage for later pages
## Summary

`n8n_manage_credentials` cannot paginate past the first 100 credentials. The `list` action returns a `nextCursor`, but there is **no input parameter to pass that cursor** (or an `offset`/`cursor` argument) back in. As a result, on any instance with more than 100 credentials, everything beyond the first page is unreachable through the tool.

## Steps to reproduce

1. Run against an n8n instance that has **> 100 credentials**.
2. Call `n8n_manage_credentials({ action: "list" })`.
3. Observe the response shape:

```json
{
  "success": true,
  "data": {
    "count": 100,
    "credentials": [ /* exactly 100 items */ ],
    "nextCursor": "eyJsaW1pdCI6MTAwLCJvZmZzZXQiOjEwMH0="
  }
}
```

4. There is no documented/available way to feed `nextCursor` back into the tool — the `list` action's input schema exposes only `action`, `data`, `id`, `includeUsage`, `name`, `type`. No `cursor` / `offset` / `limit`.

## Expected behavior

`list` should accept the `nextCursor` (and/or `limit`/`offset`) so callers can page through all credentials, mirroring how `n8n_list_workflows` accepts a `cursor`.

## Impact / knock-on bugs

This isn't just a "can't see page 2" inconvenience — it silently corrupts two other code paths:

1. **`get` by id fails for any credential beyond the first 100.** When the direct GET endpoint isn't available, `get` falls back to *list + filter*. Because the fallback only ever sees the first 100, fetching a credential whose id lives on page 2+ returns `"Credential <id> not found"` even though it exists and is actively used. This is a confusing false negative.

2. **`includeUsage` reverse-scan is incomplete.** The usage scan is built from the (capped) credential list, so credentials on later pages are missing from the results entirely — and from any `usedIn` reverse mapping derived from them. A consumer doing a credential inventory / rotation audit will silently under-report.

## Suggested fix

- Add a `cursor` (and optionally `limit`) parameter to the `list` action and thread `nextCursor` through to the n8n API, consistent with `n8n_list_workflows`.
- Have the `get` fallback paginate through all pages before concluding "not found" (or, better, hit the credential-by-id endpoint when available).
- Ensure `includeUsage` iterates all pages, not just the first.

## Environment

- Observed via the n8n-mcp MCP client against a self-hosted n8n instance with ~100+ credentials.

czlonkowski:
## Summary

Fixes #816. `n8n_manage_credentials` could not page past the first 100 credentials — the `list` action returned a `nextCursor` but exposed no `cursor`/`limit` input to feed it back. On instances with >100 credentials this also silently corrupted two other paths:

- **`get` by id** fell back to a page-1-only *list + filter*, returning a false `"Credential <id> not found"` for any id living on a later page.
- **`includeUsage`** built its inventory from the capped first page, so credentials on later pages were missing from results and any `usedIn` reverse mapping — rotation/audit consumers silently under-reported.

## Changes

- **API client**: new `listAllCredentials()` pagination helper mirroring the existing `listAllWorkflows()` (100/page, repeated-cursor guard, `MAX_PAGES=50`).
- **`list`**: threads `cursor`/`limit` to the n8n API and returns `nextCursor`, consistent with `n8n_list_workflows`.
- **`includeUsage: true`** on `list`: performs a complete all-pages scan (returns the full set, no `nextCursor`) so audits are complete.
- **`get` fallback**: scans all pages before concluding "not found".
- Tool schema (`cursor`/`limit`) and tool docs updated.

## Tests

- `listAllCredentials()`: multi-page pagination, repeated-cursor stop, `MAX_PAGES` cap.
- `handleListCredentials`: forwards `cursor`/`limit`, returns `nextCursor`, and `includeUsage` triggers a full multi-page scan with no `nextCursor`.
- `handleGetCredential` fallback: finds an id on page 2+ (regression for the false negative) and still reports "not found" for a genuinely absent id.

All touched test files green (`npm run build`, `npm run typecheck`, targeted tests pass).

## Behavior

| Call | Result |
|------|--------|
| `list` | first page + `nextCursor` |
| `list` + `cursor` | that page + `nextCursor` |
| `list` + `includeUsage:true` | ALL credentials, each with `usedIn`, no `nextCursor` |


czlonkowski:
Thanks @copilot — addressed all four review points in a5d56fa:

1. **`listCredentialsSchema` input hardening** — now mirrors `listWorkflowsSchema`: `cursor` uses `optionalEmptyAware` (empty string → `undefined`, so an echoed-back empty `nextCursor` isn't forwarded as `?cursor=`), and `limit` is bounded to `1-100`.
2. **Misleading "100 per page" tip** — reworded to "up to 100 per page (limit 1-100, default 100)" and notes the `includeUsage` scan is capped at 5000 credentials.
3. **`limit` param docs** — now state the supported `1-100` range and default, matching `n8n_list_workflows`.
4. **Tool schema text** — `includeUsage` now documents the 5000-credential cap; `limit` documents the `1-100`/default bounds.

Also applied a related code-review finding: the `list` path now strips the sensitive `data` field on both the paged and `includeUsage` paths, consistent with the existing `get`-by-id defense in depth. Added tests for limit bounds, empty-cursor normalization, and data stripping.

Issue #431, opened 2025-11-19
Codex Update Partial Workflow repeated Failure


jetchanic:
n8n version: 2.22.20

n8n-mcp version: 2.22.20

Tool: n8n_update_partial_workflow

Call body (the exact one we used with /mcp), with response:

{"success": false,
 "error": "Invalid request: request/body must NOT have additional properties",
 "code": "VALIDATION_ERROR",
 "details": {"message": "request/body must NOT have additional properties"}}

czlonkowski:
## Root Cause Identified

I've analyzed this issue and found the root cause of the `"Invalid request: request/body must NOT have additional properties"` error.

### The Problem

The error originates from `src/services/n8n-validation.ts:184-196` in the `cleanWorkflowForUpdate()` function:

```typescript
if (cleanedWorkflow.settings && typeof cleanedWorkflow.settings === 'object') {
  // Filter to only safe properties
  const filteredSettings: any = {};
  for (const key of safeSettingsProperties) {
    if (key in cleanedWorkflow.settings) {
      filteredSettings[key] = (cleanedWorkflow.settings as any)[key];
    }
  }
  cleanedWorkflow.settings = filteredSettings; // ← May be empty {}
} else {
  cleanedWorkflow.settings = {}; // ← Always sends empty object
}
```

**Two scenarios cause the error:**

1. **Line 192**: After filtering, `filteredSettings` can be an empty object `{}` if the original workflow only had non-whitelisted properties
2. **Line 195**: If no settings exist, we create an empty object `{}`

**The n8n API rejects empty settings objects** because:
- The presence of the `settings` key triggers validation
- An empty object `{}` is interpreted as "settings with no valid properties" = additional/invalid properties error

### Evidence

- **Related n8n Issue**: [#19587](https://github.com/n8n-io/n8n/issues/19587) - Same error, closed as NOT_PLANNED because it's an MCP server issue (us!)
- **Error message**: `"request/body/settings must NOT have additional properties"` - specifically mentions settings
- **Telemetry data**: Zero occurrences in production (likely because most workflows have valid settings)

### The Fix

Instead of sending empty settings `{}`, we should:

1. **Delete the settings property entirely** if it's empty after filtering
2. **Not add settings** if it didn't exist in the original workflow

```typescript
if (cleanedWorkflow.settings && typeof cleanedWorkflow.settings === 'object') {
  const filteredSettings: any = {};
  for (const key of safeSettingsProperties) {
[...]

PR #835, opened 2026-06-06
fix(validation): strip nodeGroups in cleanWorkflowForUpdate

### LINKED ISSUE #831: Bug: n8n_update_partial_workflow fails with "must NOT have additional properties" due to nodeGroups field
## Root cause
`cleanWorkflowForUpdate()` in `src/services/n8n-validation.ts` does not remove 
the `nodeGroups` field before sending the PUT request. The n8n API rejects it.

## Reproduction
# Fails:
PUT /api/v1/workflows/{id}  body: {..., "nodeGroups": []}
→ 400: {"message":"request/body must NOT have additional properties"}

# Works:
PUT /api/v1/workflows/{id}  body: {name, nodes, connections, settings}  (no nodeGroups)
→ 200 OK

## Fix
Add `nodeGroups` to the remove list in `cleanWorkflowForUpdate()`.

## Workaround
Use `n8n_update_full_workflow` instead — it only sends name/nodes/connections/settings.

## Version
n8n-mcp v2.56.0 and v2.57.1 (not fixed in latest)

paulnsorensen:
## Summary

Fixes #831.

n8n 2.x returns `nodeGroups` in `GET /workflows/{id}`, but the public API's `PUT /workflows/{id}` rejects it with:

```
400: request/body must NOT have additional properties
```

This breaks `n8n_update_partial_workflow` — including `activateWorkflow` / `deactivateWorkflow` operations — against n8n 2.x instances (reproduced on n8n 2.23.2 with n8n-mcp 2.57.1).

## Change

- Add `nodeGroups` to the read-only field strip list in `cleanWorkflowForUpdate()` (`src/services/n8n-validation.ts`), matching the existing pattern for `versionCounter`, `description`, etc.
- Add a unit test mirroring the existing Issue #431 / versionCounter coverage.

## Testing

- `npx vitest run tests/unit/services/n8n-validation.test.ts` — 105 passed, 0 failed
- `tsc --noEmit` — clean
- Reproduced the original failure against a live n8n 2.23.2 instance via `n8n_update_partial_workflow` `activateWorkflow`; the PUT body minus `nodeGroups` is accepted (per the reproduction in #831).

### DIFF
--- a/src/services/n8n-validation.ts
+++ b/src/services/n8n-validation.ts
@@ -172,6 +172,7 @@ export function cleanWorkflowForUpdate(workflow: Workflow): Partial<Workflow> {
     active,
     activeVersionId,
     activeVersion,
+    nodeGroups, // n8n 2.x returns this but rejects it in PUT (Issue #831)
     // Keep everything else
     ...cleanedWorkflow
   } = workflow as any;
--- a/tests/unit/services/n8n-validation.test.ts
+++ b/tests/unit/services/n8n-validation.test.ts
@@ -429,6 +429,22 @@ describe('n8n-validation', () => {
         expect(cleaned.name).toBe('Test Workflow');
       });
 
+      it('should exclude nodeGroups field for n8n 2.x API compatibility (Issue #831)', () => {
+        const workflow = {
+          name: 'Test Workflow',
+          nodes: [],
+          connections: {},
+          versionId: 'v123',
+          nodeGroups: [], // n8n 2.x returns this but rejects it in PUT
+        } as any;
+
+        const cleaned = cleanWorkflowForUpdate(workflow);
+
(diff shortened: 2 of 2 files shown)


czlonkowski:
Closing as superseded by #840, which shipped in **v2.57.2**. Rather than adding `nodeGroups` to the denylist, we inverted `cleanWorkflowForUpdate()` to an allowlist (`name`, `nodes`, `connections`, `settings` only) so this whole class of bug — n8n adding new top-level fields to the GET response that its PUT schema rejects — can't recur. The regression test in #840 covers the `nodeGroups` echo-back case your PR addressed.

Thank you for jumping on this so quickly — sorry we couldn't merge it directly, and your contribution is appreciated!

PR #839, opened 2026-06-07
fix: strip `nodeGroups` in cleanWorkflowForUpdate to prevent 400 on update

### LINKED ISSUE #838: cleanWorkflowForUpdate leaks `nodeGroups` → every update 400s with "request/body must NOT have additional properties"
## Summary
On n8n instances recent enough to include the **node-grouping / workflow-versioning** feature, the workflow GET response contains a top-level `nodeGroups` field. `cleanWorkflowForUpdate()` strips every other read-only/new field but **not `nodeGroups`**, so it survives into the `PUT /workflows/{id}` body. The n8n public API validates that body against the `workflow` schema, which is `additionalProperties: false` and does not list `nodeGroups` → the request is rejected with:

```
400 request/body must NOT have additional properties
```

This makes **100% of write operations fail** (`n8n_update_partial_workflow`, `n8n_update_full_workflow`, `n8n_autofix_workflow`, version restore) on affected instances. Reads are unaffected.

## Environment
- n8n-mcp: **2.57.1** (the relevant code is identical on `main`)
- n8n: recent build exposing workflow versioning (`activeVersion`, `activeVersionId`, `versionCounter`) and `nodeGroups`; public API spec version `1.1.1`

## Root cause
`src/services/n8n-validation.ts` → `cleanWorkflowForUpdate()` uses a **denylist** destructure that omits `nodeGroups`:

```ts
const {
  id, createdAt, updatedAt, versionId, versionCounter, meta, staticData,
  pinData, tags, description, isArchived, usedCredentials, sharedWithProjects,
  triggerCount, shared, active, activeVersionId, activeVersion,
  ...cleanedWorkflow            // <-- nodeGroups is NOT destructured, so it stays
} = workflow as any;
```

The instance's own OpenAPI (`/api/v1/openapi.yml`) defines `workflow` with `additionalProperties: false`; `nodeGroups` is not an allowed property, so the PUT 400s.

## Reproduction (verified against a live instance)
Same body, single variable toggled:

| PUT body | Result |
|---|---|
| `{name, nodes, connections, settings}` **+ `nodeGroups`** | `400 request/body must NOT have additional properties` |
| `{name, nodes, connections, settings}` (no `nodeGroups`) | `200 OK` |

Note: the settings filter already correctly drops the newer `binaryMode` key — `nodeGroups` is the only remaining leak.

## Fix (minimal)
Add `nodeGroups` to the destructure so it is stripped before PUT. PR opened against `main`.

## Recommended deeper fix (robustness)
The denylist breaks every time n8n adds a new top-level field (this is at least the second such field after the versioning set). Consider inverting to an **allowlist** — keep only the fields the public API accepts (`name, nodes, connections, settings, staticData`) — so future n8n field additions can't reintroduce this class of bug.

## Note for n8n core
n8n's GET returns `nodeGroups` (and `activeVersionId`, `versionCounter`) but its public-API `workflow`/PUT schema is `additionalProperties: false` and omits them — i.e., the API returns fields its own update endpoint rejects. Adding these to the schema would let updates round-trip node-group data instead of silently dropping it.

schaferjart:
## What
Adds `nodeGroups` to the destructure in `cleanWorkflowForUpdate()` so it is stripped before the `PUT /workflows/{id}` request.

## Why
Recent n8n versions return a top-level `nodeGroups` field on workflows (alongside the `activeVersion`/`versionCounter` versioning fields). `cleanWorkflowForUpdate()` strips every other read-only/computed field but not `nodeGroups`, so it leaked into the update body. The n8n public API validates the body against the `workflow` schema (`additionalProperties: false`), which does not list `nodeGroups`, and rejects it with:

```
400 request/body must NOT have additional properties
```

This breaks **100% of write operations** (`n8n_update_partial_workflow`, `n8n_update_full_workflow`, `n8n_autofix_workflow`, version restore) on affected instances.

Fixes #838.

## Reproduction (verified against a live instance)
| PUT body | Result |
|---|---|
| `{name, nodes, connections, settings}` **+ `nodeGroups`** | `400 ... must NOT have additional properties` |
| `{name, nodes, connections, settings}` (no `nodeGroups`) | `200 OK` |

After this patch, `cleanWorkflowForUpdate({…, nodeGroups: [], versionCounter: 9, activeVersion: {…}})` returns only `{name, nodes, connections, settings}` and updates succeed end-to-end.

## Diff
```diff
     active,
     activeVersionId,
     activeVersion,
+    nodeGroups,
     // Keep everything else
     ...cleanedWorkflow
   } = workflow as any;
```

## Note / possible follow-up
This is the minimal fix. The denylist approach breaks whenever n8n adds a new top-level field (this is the second occurrence after the versioning set). A more durable fix would invert to an **allowlist** — keep only `{name, nodes, connections, settings, staticData}` — so future n8n field additions can't reintroduce this class of bug. Happy to follow up with that if preferred.

### DIFF
--- a/src/services/n8n-validation.ts
+++ b/src/services/n8n-validation.ts
@@ -172,6 +172,7 @@ export function cleanWorkflowForUpdate(workflow: Workflow): Partial<Workflow> {
     active,
     activeVersionId,
     activeVersion,
+    nodeGroups,
     // Keep everything else
     ...cleanedWorkflow
   } = workflow as any;


czlonkowski:
Closing as superseded by #840, which shipped in **v2.57.2**. Instead of extending the denylist we inverted `cleanWorkflowForUpdate()` to an allowlist (`name`, `nodes`, `connections`, `settings` only), so `nodeGroups` and any future echo-back fields n8n adds are stripped without needing a patch each time.

Thank you for the fix and especially for the excellent root-cause analysis in #838 — your allowlist recommendation there is exactly what we shipped, and the test in #840 credits the reports that drove it. Much appreciated!

PR #840, opened 2026-06-09
fix(workflow): allowlist write payload so n8n 2.x echo-back fields don't break updates

czlonkowski:
## Problem

A community user (Luca) reported that `n8n_update_partial_workflow` consistently fails against self-hosted n8n 2.x with:

```
Invalid request: request/body must NOT have additional properties
```

`validateOnly: true` passes, but the real apply fails for **every** operation type. Their workaround was to strip `availableInMCP` from settings and switch to `n8n_update_full_workflow` (making `name` required, which the tool docs list as optional).

## Root cause

n8n's Public API write schema (`workflow.yml`, used for `PUT /workflows/{id}`) declares `additionalProperties: false`. But n8n's `GET` response echoes back server-managed fields that the write schema does **not** accept — including a **top-level `availableInMCP`** column on n8n 2.x, plus fields not even in the OpenAPI spec (`activeVersionId`, `versionCounter`, confirmed against a live instance).

`cleanWorkflowForUpdate` used a **denylist** (drop known read-only fields). Any field n8n newly echoes that the denylist doesn't explicitly drop leaks into the write payload → `request/body must NOT have additional properties`. The error is at the **top level** (no `/settings` subpath), which is why this differs from the older settings-only reports — `availableInMCP` *inside* `settings` is actually a valid writable property on 2.23.x.

I verified the n8n@2.23.4 public-API schemas directly: the write body accepts only `name`, `description`, `nodes`, `connections`, `settings`, `staticData`, `pinData`; everything else is read-only/rejected.

## Fix

1. **`cleanWorkflowForUpdate` → allowlist.** Keep only the writable top-level fields the previous denylist already effectively sent (`name`, `nodes`, `connections`, `settings`); `settings` is still filtered to its own writable allowlist (which keeps `availableInMCP`). This is **forward-compatible** — any field n8n adds in future versions can never break updates again. Fixes `n8n_update_partial_workflow`.
2. **`handleUpdateWorkflow` always fetches + merges.** n8n's full-replace `PUT` requires `name`, `nodes`, `connections` and `settings` to all be present. The handler now always fetches the current workflow and merges the caller's partial update over it, so omitting `name`/`settings` no longer fails with `request/body must have required property 'name'`. Fixes Luca's `n8n_update_full_workflow` finding.

## Verification

- `npm run typecheck` clean; all affected unit suites pass (**215** in the two touched files, **406** across all affected suites).
- New regression tests:
  - allowlist strips unknown top-level fields (`availableInMCP`, `activeVersionId`, and a hypothetical future field) — proving forward-compatibility a denylist can't guarantee;
  - top-level `availableInMCP` stripped while the in-`settings` one is preserved;
  - full update omitting `name`/`settings` preserves them from the current workflow;
  - settings-only update (no graph change) still fetches + merges.
- Confirmed end-to-end against a live n8n instance: partial update applies; full update omitting `name`/`settings` preserves both (the instance returns top-level `activeVersionId`/`versionCounter`, which the allowlist now strips).

Thanks to Luca for the clear report with the diagnostic matrix and reproduction steps.


czlonkowski:
### Supersedes #835 and #839 (and resolves #831, #838)

Both open PRs #835 and #839 fix the same `nodeGroups` symptom by adding it to the **denylist** in `cleanWorkflowForUpdate`. This PR fixes that and the whole class of bug by inverting to an **allowlist** (`name`, `nodes`, `connections`, `settings`), so `nodeGroups` — and any future top-level field n8n starts echoing on read — is stripped automatically. This is exactly the durable allowlist follow-up @schaferjart suggested in #839.

Added explicit `nodeGroups` coverage to the allowlist regression test, so the linked issues (#831, #838) are demonstrably covered.

Credit to @schaferjart and @paulnsorensen for the independent reproductions on live 2.23.x instances.

PR #850, opened 2026-06-19
chore(deps): clear shipped Dependabot advisory (uuid) + dev/build tooling

czlonkowski:
## Summary

Triage of the **72 open Dependabot alerts** (4 critical / 28 high / 33 moderate / 7 low) found that almost none reach a shipped artifact. The published npm package and Docker runtime image install from **`package.runtime.json`** (not the full `package.json`) plus a pre-built `data/nodes.db` — the n8n packages (`n8n-core`, `n8n-nodes-base`, `n8n-workflow`, `@n8n/n8n-nodes-langchain`) are **build-time only** and never ship.

A clean-room `npm install --production` from `package.runtime.json` flags **exactly one** vulnerable package: **`uuid@10.0.0`**. This PR clears it, plus the dev/build-tooling advisories.

## Changes

| Dependency | From → To | Scope | Clears |
|---|---|---|---|
| `uuid` | `^10.0.0` → `^14.0.0` (`package.json` + `package.runtime.json`) | **runtime (shipped)** | GHSA-w5hq-g745-h8pq / CVE-2026-41907 |
| `vitest` + `@vitest/*` | `^3.2.4` → `^3.2.6` | dev | CVE-2026-47429 (+ refreshes `vite` → 7.3.5) |
| `ui-apps` `vite` | `^6.0.0` → `^6.4.3` | dev/build | ui-apps `vite`/`rollup`/`postcss`/`@babel/core`/`picomatch` (refreshed to 4.62.0 / 8.5.15 / 7.29.7 / 2.3.2+4.0.4) |

`uuid` is the only advisory that reaches the runtime artifact. The server's own usage is unaffected — `v4` (random) everywhere, plus one `v5(name, namespace)` call that passes no output buffer, so the bounds-check bug isn't reachable — but it's bumped to clear the advisory. `v4`/`v5` named exports are unchanged and verified under the CommonJS build.

## Verification

- ✅ `npm run typecheck`
- ✅ `npm run build`
- ✅ `ui-apps` build (vite 6.4.3, all 5 single-file apps)
- ✅ Full unit suite — **150 files, 4944 tests passed**, 35 skipped

## Not addressed (intentionally)

- **~59 n8n-transitive advisories** (`fast-xml-parser`, `tar`, `thrift`, `@grpc/grpc-js`, `@opentelemetry/*`, `protobufjs`, `nodemailer`, etc.) are **build-time only** — absent from the runtime install. They clear on the regular `npm run update:n8n` cadence.
- **One low-severity `esbuild`** advisory (GHSA-g7r4-m6w7-qqqr — Windows dev-server arbitrary file read) stays pinned by `vite@7`'s `^0.27.0` range. It's unreachable here (esbuild is an internal Vite/Vitest transform, not an exposed dev server; CI is Linux). Clearing it would require the `vite 8` / `vitest 4` major migration, out of scope for this security bump.


czlonkowski:
Thanks — the observation is factually correct, and I verified it: the root `package-lock.json` still has **22 vulnerable `uuid` installs** (<11.1.1) — `8.3.2` ×4, `9.0.1` ×6, `10.0.0` ×12.

The key distinction is **which uuid ships**. Every one of those 22 is a build-time-only transitive of the n8n packages:

- the `10.0.0` copies are all under `@langchain/*` (pulled by `@n8n/n8n-nodes-langchain`)
- the `8.x`/`9.x` copies are under `snowflake-sdk`, `tedious`, `@n8n/typeorm`, `@azure/msal-node`, `google-gax`, `gaxios`, `weaviate-client`, etc.

None of these are in `package.runtime.json`, which is what the npm package and the Docker runtime stage actually install. A clean-room `npm install --production` from `package.runtime.json` resolves the **direct** `uuid` to `14.0.0` and contains **no other uuid** — so the shipped artifact has no vulnerable uuid. The n8n packages themselves are build-time-only here (used to generate `data/nodes.db`) and never ship.

I deliberately did **not** add a global `overrides: { uuid }` to force those transitives to 14: that would push `uuid` v8/v9 consumers (typeorm, snowflake-sdk, tedious, …) across multiple breaking majors and risks failing `npm run rebuild`, all for code that never reaches production. This is the same posture as the other ~59 n8n-transitive Dependabot advisories — they clear when the n8n packages bump upstream (on the regular `npm run update:n8n`), not here.

So to be precise about scope: this PR removes vulnerable `uuid` from the **shipped runtime artifact** (the real attack surface). The GitHub `uuid` alert may remain open until n8n bumps its own `uuid`, because Dependabot scans the full build lockfile. I've tightened the CHANGELOG wording (96a5b38) to say exactly that rather than "clears the advisory."

anpe-efficy:
Heads up — this bump broke CommonJS consumers.

`uuid@14` is ESM-only, but the compiled output still calls `require('uuid')` (e.g. `dist/services/workflow-auto-fixer.js`). When the published package runs under Node's CommonJS loader it crashes on startup:

```
Error [ERR_REQUIRE_ESM]: require() of ES Module .../node_modules/uuid/dist-node/index.js from .../n8n-mcp/dist/services/workflow-auto-fixer.js not supported.
```

This hits every release since 2.59.1 (2.59.1, 2.59.2, 2.59.3). The process dies right away, so MCP clients just see `Connection closed`.

Repro: `npx n8n-mcp@2.59.3`. 2.59.0 still works (uuid@^10).

The unit suite passed because it doesn't exercise that require path against the ESM-only build. Since the advisory this cleared isn't reachable anyway (as you noted above), pinning uuid back to `^11` would fix it and keep a CommonJS build — uuid's own deprecation notice recommends 11 for CJS codebases.

Filed #864 with the full details.

PR #874, opened 2026-06-24
chore(deps): harden Dependabot config (ignore n8n pkgs, cover ui-apps)

czlonkowski:
Follow-up to #867. A Codex review of the Dependabot PRs that #867's config generated (#868–#873) surfaced three coverage/safety gaps in `.github/dependabot.yml`:

### Changes
1. **Ignore the n8n packages** (`n8n`, `n8n-core`, `n8n-workflow`, `n8n-nodes-base`, `@n8n/*`).
   These are updated via `npm run update:n8n`, which also rebuilds `data/nodes.db` while preserving community nodes (`scripts/update-n8n-deps.js`). A plain Dependabot bump would merge a **stale node catalog** — so they're excluded here and stay on the dedicated flow.
2. **Add a `/ui-apps` npm update block.** `ui-apps/` has its own `package-lock.json` (React/Vite, built by `npm run build:ui`) and is not a root workspace, so Dependabot wasn't monitoring it.
3. **Document `package.runtime.json`.** The Docker/npm runtime images install from this non-standard manifest, which Dependabot can't manage directly — added a note to mirror runtime-dep bumps into it.

### Why
Without (1), Dependabot would soon open `n8n*` PRs that pass CI but ship an out-of-date node database. (2) and (3) close real coverage holes so UI and runtime dependencies actually receive security updates.

Validated as well-formed YAML (4 update blocks: `npm:/`, `npm:/ui-apps`, `github-actions:/`, `docker:/`).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,12 @@
 
 version: 2
 updates:
-  # Monitor npm dependencies
+  # Monitor root npm dependencies
+  #
+  # NOTE: package.runtime.json (the slim manifest the Docker/npm runtime images
+  # install from) is a non-standard file Dependabot cannot manage directly. When
+  # bumping a runtime dependency here, mirror the change into package.runtime.json
+  # so the published images pick it up.
   - package-ecosystem: "npm"
     directory: "/"
     schedule:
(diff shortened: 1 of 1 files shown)


czlonkowski:
Addressed both Copilot comments:

1. **`@n8n/*` wildcard** — Dependabot's `ignore.dependency-name` *does* support `*` globs (per the [options reference](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference), e.g. `react*`, `org.xwiki.*`), so the wildcard works. That said, I've added `@n8n/n8n-nodes-langchain` explicitly alongside `@n8n/*` so the intent is unambiguous for the package we actually depend on (the wildcard stays to cover future `@n8n/*` packages).
2. **`/ui-apps` update types** — good catch on the inconsistency; the `/ui-apps` block now has the same `production`/`development` minor+patch `groups` as the root block, so UI updates are batched the same way instead of one PR per dependency.

(For both blocks, major updates still get individual PRs — matching the root policy. If you'd prefer to also suppress major auto-PRs for the UI build toolchain specifically, that's a one-line `ignore: [update-types: version-update:semver-major]` opt-in — say the word.)

PR #908, opened 2026-07-03
fix: stale hardcoded version in health check, deprecated API, typos

fix2015:
A few things I spotted:

- \`healthCheck()\` was returning a hardcoded version \`'2.24.1'\` in both the healthy and unhealthy branches, but package.json is at 2.62.0. Imported \`PROJECT_VERSION\` from \`utils/version\` instead.
- \`res.finished\` was deprecated in Node 13.4.0 — changed to \`res.writableEnded\`
- fixed "paralel" and "Concieved" typos in CLAUDE.md
- replaced the placeholder \`issues/XXX\` link in shared-database.ts with an inline description
- removed an unnecessary \`(this as any)\` cast in workflow-validator — \`currentWorkflow\` is already typed as a private field on the class


czlonkowski:
Thanks @fix2015, going in via #1059 with your commit carried over (the CLAUDE.md typo hunk was dropped, already fixed on main). A test now pins the health check to the package version. Closing this one in favour of #1059; you are credited there.

Issue #903, opened 2026-07-02
chore: dependency security audit on the default branch


czlonkowski:
## Context

Automated dependency scanning has accumulated a backlog of alerts on the default branch. For a project positioning itself around production reliability (Discussion #900), dependency hygiene is part of the product story, not just maintenance.

## Tasks

- [ ] Triage all open Dependabot alerts: classify by runtime vs dev/build-only dependency and by actual exploitability in this codebase (many alerts in transitive dev dependencies have no runtime exposure).
- [ ] Upgrade or pin runtime dependencies with applicable advisories first.
- [ ] Batch dev-dependency updates where majors allow; dismiss non-applicable alerts with a stated reason.
- [ ] Verify `npm run build`, `npm test`, and `npm run test:integration` pass after upgrades.
- [ ] Consider enabling Dependabot grouped security PRs to keep the backlog from re-accumulating.

## Acceptance

Zero open critical/high alerts that affect runtime code paths; remaining alerts dismissed with documented reasoning.

czlonkowski:
## Audit complete

**Baseline:** 67 open Dependabot alerts (3 critical / 27 high / 29 medium / 8 low).

### Runtime-reachable → fixed (PR #906)
- **`form-data` CRLF injection** (`<4.0.6`). The form-trigger handler (`src/triggers/handlers/form-handler.ts`) puts request-supplied filenames/field names into a multipart body on the `/form/<webhookId>` path. Bumped the direct dependency `^4.0.5 → ^4.0.6`. Build green, form-handler suite 48/48.

### Everything else → not runtime-reachable
All remaining alerts originate in the **n8n node-package graph** (`@n8n/n8n-nodes-langchain`, `n8n-nodes-base`, `n8n-core`, `n8n-workflow`) — AWS SDK (`fast-xml-parser`), Milvus (`thrift`), gRPC, IMAP (`utf7`/`semver`), `nodemailer`, `tar`, `hono`, `undici`, etc. n8n-mcp loads these only to introspect node **metadata** during `npm run rebuild`; the server never executes those nodes at runtime, so the vulnerable code paths are unreachable. These four packages are already Dependabot-`ignore`d and clear via `npm run update:n8n`.

Two version-match-but-path-not-reached cases documented individually:
- **`qs`** — express consumes `qs.parse` (req.query), not the affected `qs.stringify` comma/`encodeValuesOnly` path; we never call qs directly.
[...]

PR #912, opened 2026-07-03
ci: skip live n8n integration tests when no instance is configured

czlonkowski:
## Problem

The `test` job's integration step talks to a live n8n instance via the `N8N_API_*` secrets. Those secrets are unavailable on **Dependabot and fork PRs**, so the live suites fail outright and the required `test` check goes **permanently red** — the dependency PRs we most want to auto-merge can never turn green. A couple of live-API cases also flake intermittently even when the instance is up.

## Change

Split the integration run and gate only the live part:

- **Offline suites** (database, security, mcp-protocol, workflow-diff, templates, docker, `flexible-instance-config`, …) — run in their own step that **always executes and stays a blocking gate on every PR**, Dependabot/forks included. They need no secrets (verified green with none set).
- **Live suites** (`tests/integration/n8n-api/**`, `tests/integration/ai-validation/**`) — run in a separate step gated on a job-level boolean `HAS_N8N_INSTANCE`. It's derived from **both** `N8N_API_URL` and `N8N_API_KEY` (the live tests' `getN8nCredentials()` throws if either is missing), so a partial config won't run-and-fail. Skipped when no instance is configured; blocking on in-repo PRs where the secrets exist. The `N8N_API_*` secrets are scoped to this step only, not the whole job.

Unit tests, lint, and typecheck remain always-on blocking gates.

## Why not blanket `continue-on-error`

That would drop the gate for *every* PR, including the offline suites that catch real regressions. This keeps the offline suites gating everywhere and only relaxes the genuinely instance-dependent tests where they can't run.

## Notes

- No `--retry` is added: the base Vitest config sets `retry: 0` on purpose ("fix flaky tests, don't mask them"). The two flaky live-API cases in `update-partial-workflow.test.ts` (moveNode/disableNode) should be fixed with proper eventual-consistency waits — tracked as a follow-up.

Refs #903

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,10 +42,11 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 15  # Increased from 10 to accommodate larger database with community nodes
     env:
-      # Surfaced at job level so the integration step's `if:` can detect whether a
-      # live n8n instance is configured. The `secrets` context is not available in
-      # `if:` conditions, but `env` is — so we route the URL through env here.
-      N8N_API_URL: ${{ secrets.N8N_API_URL }}
+      # A boolean flag (not the secret itself) so the live-integration step's `if:`
+      # can detect whether an instance is configured without exposing the URL to
+      # every step. `secrets` is not usable in `if:`, but `env` is; evaluating the
+      # secret to a bool here keeps the actual value scoped to the step that needs it.
+      HAS_N8N_INSTANCE: ${{ secrets.N8N_API_URL != '' }}
     steps:
       - uses: actions/checkout@v7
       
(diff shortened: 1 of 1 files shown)


czlonkowski:
Thanks — addressed the review in 0e4716c:

1. **Secret exposure at job level** — fixed. The job now exposes only a boolean `HAS_N8N_INSTANCE: ${{ secrets.N8N_API_URL != '' }}`; the actual `N8N_API_*` secrets are scoped to the live-integration step only.

2. **Gate on the flag, scope secret to the step** — done, as above.

3. **Skipping *all* integration suites contradicts the rationale** — good catch, fixed properly. The live-instance tests are cleanly contained in `tests/integration/n8n-api/**` and `tests/integration/ai-validation/**`, so I split the run:
   - **Offline suites** (database, security, mcp-protocol, workflow-diff, templates, docker, `flexible-instance-config`, …) — always run, blocking on every PR including Dependabot/forks (verified they pass with no secrets).
   - **Live n8n-api / ai-validation** — gated on the flag, skipped only when no instance is configured.

   Offline-integration coverage is now preserved on secretless PRs, consistent with the "not blanket `continue-on-error`" rationale.

4. **`--retry` contradicts the repo's `retry: 0` philosophy** — agreed, removed it entirely rather than encoding it in config. The two flaky live-API cases (`update-partial-workflow.test.ts` moveNode/disableNode, "expected false to be true") look like eventual-consistency timing and should be fixed with proper waits — tracking as a separate follow-up rather than masking with retries.

czlonkowski:
Addressed the follow-up round in ad8a139 + a description update:

1. **`HAS_N8N_INSTANCE` derived only from the URL** — fixed. It now requires **both** `N8N_API_URL` and `N8N_API_KEY` (`${{ secrets.N8N_API_URL != '' && secrets.N8N_API_KEY != '' }}`). Confirmed `getN8nCredentials()` throws if either is missing, so a partially-configured instance would otherwise run-and-fail. Good catch.

2. **Comment about `secrets` in `if:`** — reworded. I'm gating on a job-level `env` flag rather than referencing `secrets` in the step `if:` because, per GitHub's [context-availability table](https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability), `secrets` isn't available in a step `if:` — but the flag's *main* purpose is to avoid exposing the secret values to every step (your earlier point), so I rewrote the comment to lead with that and note the secrets are empty on Dependabot/fork PRs, per your suggestion.

3. **`--retry=2` mentioned but not in the command** — the retry was removed intentionally in the previous round (it conflicts with the repo's `retry: 0` "fix, don't mask" policy). The stale reference was in the PR description, not the code — I've updated the description to drop it. The two flaky live-API tests are tracked as a follow-up to fix properly rather than retry.

PR #927, opened 2026-07-10
fix: use whitelist approach in cleanWorkflowForUpdate to prevent API errors

manato-tajiri:
## Problem

The  function used a blacklist approach that destructured out specific known properties but passed through any unknown properties via the rest spread. n8n cloud instances return properties like , , , , ,  that were not in the blacklist, causing the n8n API to reject requests with "additional properties" errors.

This affects all  operations (updateName, updateNode, addNode, addConnection, removeConnection).

## Fix

Changed  from a **blacklist** approach to a **whitelist** approach. The function now only includes the 6 properties the n8n API accepts in :
- 
- 
- 
- 
- 
- 

## Testing

- All 76 existing unit tests pass
- Build succeeds with 
> n8n-mcp@2.28.3 build
> tsc -p tsconfig.build.json

Fixes the "additional properties" error on all partial update operations for n8n cloud instances.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "n8n-mcp",
-  "version": "2.28.2",
+  "version": "2.28.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "n8n-mcp",
-      "version": "2.28.2",
+      "version": "2.28.3",
       "license": "MIT",
       "dependencies": {
         "@modelcontextprotocol/sdk": "1.20.1",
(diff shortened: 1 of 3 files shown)


manato-tajiri:
Note: The current npm version (2.63.2) already has a whitelist approach but is missing  and . This PR adds those missing properties to the whitelist.

czlonkowski:
Thanks. Since this was opened, `cleanWorkflowForUpdate` moved to an explicit allowlist (`name`, `nodes`, `connections`, `nodeGroups`, `settings`, `parentFolderId`) that deliberately leaves `active` and `tags` out of update payloads. Activation and tags are managed through their own endpoints, and n8n's PUT rejects `active`. If a current release still fails against n8n Cloud for you, please open an issue with the response body. Closing as superseded.

PR #943, opened 2026-07-21
feat(http): add stateless dual-era MCP transport with SDK v2

LevSky22:
## Summary

- add request-scoped stateless Streamable HTTP mode for remote MCP clients
- move the server side to the split MCP TypeScript SDK v2 packages
- serve legacy initialize-era clients and MCP 2026-07-28 clients from the same endpoint
- preserve tenant isolation, CORS handling, disabled-tool policy, and n8n-mode tool descriptions
- retain SDK v1 only for the embedded outgoing WebSocket MCP client, because SDK v2 does not provide that transport

## Compatibility

Existing legacy clients continue to initialize without a session ID. SDK v2 clients can negotiate the modern protocol and use the same tool surface. The implementation remains stateless: each HTTP request receives an isolated server/transport pair.

This PR is transport-only. Response bounding and artifact pagination are intentionally split into a separate independent PR.

## Verification

- `npm run build`
- 82 focused transport and compatibility tests:
  - `tests/integration/mcp-protocol/dual-era-stateless.test.ts`
  - `tests/unit/http-server-session-management.test.ts`
  - `tests/unit/http-server-n8n-mode.test.ts`
- production Docker image build
- rebased onto upstream `main` at 2.68.1

The branch also declares the SDK v2 client/core packages used by its integration test as development-only dependencies.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -47,6 +47,12 @@ MCP_MODE=stdio
 PORT=3000
 HOST=0.0.0.0
 
+# Streamable HTTP transport lifecycle
+# - stateful (default): issue MCP session IDs and retain per-session transports
+# - stateless: process each POST independently; no session IDs or resumability
+# Stateless mode is recommended for API-style and horizontally scaled clients.
+# MCP_HTTP_TRANSPORT_MODE=stateful
+
 # Base URL Configuration (optional)
 # Set this when running behind a proxy or when the server is accessed via a different URL
 # than what it binds to. If not set, URLs will be auto-detected from proxy headers (if TRUST_PROXY is set)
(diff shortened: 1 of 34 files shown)


czlonkowski:
Thanks for the work here. The problem this targeted, 2026-07-28-era clients probing with a session-less `server/discover` and getting a session error back, was fixed on main with a much smaller change: unknown JSON-RPC methods now return `-32601` before any session handling (see the CHANGELOG entry after 2.66.3). The SDK stays on 1.x for now; moving to the split v2 packages will be its own effort once they settle. Closing as superseded.

Issue #774, opened 2026-05-05
`n8n_list_workflows` and `n8n_executions` fail on first page due to empty string optional parameters


ale90bsas:
**Environment**:
- n8n-mcp v2.50.3
- n8n v2.18.6
- opencode v1.14.35 (MCP client)

**Description**:

When called from opencode, the MCP client serializes all schema fields into the JSON payload — including optional ones — as empty strings (`""`) or empty arrays (`[]`) rather than omitting them. The handler passes these values directly to the n8n API, which rejects them with `VALIDATION_ERROR`.

**Affected tools**: `n8n_list_workflows`, `n8n_executions`

**Errors observed**:
- `Empty value found for query parameter 'cursor'`
- `Empty value found for query parameter 'projectId'`
- `Empty value found for query parameter 'workflowId'`
- `An invalid cursor was provided` (when passing placeholder like `"0"`)

**Root cause**:

In `handlers-n8n-manager.js`, optional parameters are passed through to the API client without filtering falsy values:

```js
// Before
cursor: input.cursor,
projectId: input.projectId,
workflowId: input.workflowId,
status: input.status,
```

When the client sends `cursor: ""`, axios serializes it as `?cursor=` in the query string, which the n8n API rejects.

**Fix applied locally**:

```js
// After
cursor: input.cursor || undefined,
projectId: input.projectId || undefined,
workflowId: input.workflowId || undefined,
status: input.status || undefined,
```

This ensures empty strings are coerced to `undefined` and omitted by axios when building the query string.

**Lines patched**:
- `handleListWorkflows` — `cursor` (line ~710), `projectId` (line ~713)
- `handleListExecutions` — `cursor` (line ~1170), `workflowId`, `projectId`, `status`

**Note**: `cursor` in `handleListWorkflows` already had `|| undefined` from a previous fix attempt but `projectId` did not, causing the error to shift from one parameter to the next.

**Attempted workaround via prompt instructions**:

Before patching the handler, an attempt was made to work around the issue using [[n8n-skills](https://github.com/czlonkowski/n8n-skills/tree/main)](https://github.com/czlonkowski/n8n-skills/tree/main). The skill included an explicit rule:

> When calling `n8n_list_workflows` or `n8n_executions` for the first page, omit ALL optional parameters entirely. Do NOT pass empty strings, empty arrays, or placeholder values.

[...]

PR #948, opened 2026-07-25
fix: preserve and author n8n canvas groups (nodeGroups)

czlonkowski:
## The bug

n8n 2.28 added **canvas groups** (Cmd+G), stored on the workflow as `nodeGroups: [{id, name, nodeIds, description?}]`. n8n validates them on **every** write — including writes that have nothing to do with grouping.

Our update payload dropped the field, so n8n backfilled the *stored* groups and validated them against the graph we submitted. Reproduced live against a 2.30 instance:

| write | result |
|---|---|
| edit a param on a grouped node | 200, groups kept |
| remove a grouped node | **400** `Group "X" references node ID "…" that does not exist in the workflow.` |
| insert a node between two grouped nodes | **400** `Node group "X" (…) must form a single connected subgraph with a single entry and exit.` |

So `n8n_update_partial_workflow`, `n8n_update_full_workflow`, autofix, rollback and version restore all failed on any grouped workflow as soon as a diff touched a group. As Cmd+G adoption grows this becomes a steady stream of "MCP can't edit my workflow".

## The fix

Groups are carried through writes and reconciled with the finished graph **once**, in a `finalizeNodeGroups` hook called before every success path (including `validateOnly`): members a diff deleted are pruned, a group left empty is removed. Reconciling at the end rather than inside `applyRemoveNode` means a node removed and re-added in the same batch keeps its membership. Nodes and connections are never altered to save a frame; every adjustment is reported as a warning.

**Topology stays n8n's decision.** Its rejection names the offending group, so the server adjudicates and we ungroup precisely that one — no local copy of n8n's grouping rules to drift from the running instance. The shared validator combines an all-connection-type connectivity search with main-only entry/exit extraction and node-type metadata for triggers; a vendored copy would delete groups the server would have accepted.

**Authored vs inherited.** Groups the caller supplied in the same request are never silently discarded — n8n's message is surfaced. Groups that merely came back from a GET degrade with a warning so the unrelated edit still lands.

**Capability is discovered from the write, not a version probe** (`/rest/settings` exposes no version on many instances). An instance that rejects the field gets the workflow without it; one that rejects group descriptions (n8n 2.32+) gets the groups without them. Both warn. The verdict is remembered per client — but only for *schema* rejections, so one invalid group can't disable grouping for the whole instance.

No new runtime dependency: `package.runtime.json` ships 12 packages and `n8n-workflow` is not among them, so importing n8n's validator would throw `MODULE_NOT_FOUND` for every Docker user.

## Added

- **`setNodeGroups`** diff operation — replaces the canvas groups, members by name or ID, `[]` ungroups everything.
- **`nodeGroups`** on `n8n_create_workflow` and `n8n_update_full_workflow`. Absent = keep stored groups, `[]` = ungroup all, `null` = absent (clients that serialize every field, #774).
- Groups in `n8n_get_workflow` `structure` / `filtered` (groups touching the requested nodes) / `active` (the **published** version's groups, not the draft's), and in `n8n_deploy_template`.
- `validate_workflow` warns about dangling members, empty groups, a node in two groups, and a trigger inside a group — always warnings, never blocking.

## Two fixes found along the way

- **`updateNode` no longer accepts an `id` rewrite.** It silently orphaned canvas-group membership and pinned data.
- **`compareVersions` was blind to `nodeGroups`,** so a group-only version read as identical to its predecessor.

## Verification

- `npm run typecheck` clean; **5394 unit tests pass**, 81 new.
- New `tests/unit/services/node-groups.test.ts` (31): repair, sanitization, error classification against the two real n8n messages, deep-frozen input (rollback payloads and version snapshots are shared objects).
- `n8n-api-client.test.ts` (+13): the whole ladder against a mocked client, including old-instance behaviour — the correct substitute for a pre-2.28 instance we don't have.
- `workflow-diff-engine.test.ts` (+22), `handlers-*.test.ts` (+11), `workflow-validator.test.ts` (+5): op validation, prune/re-add, rename safety, id immutability, merge trio, op-schema survival of a stringified payload.
- Live scenarios replayed against a real n8n instance (create with a group, param edit, removeNode, insert-between, authoring, ungroup-all, description strip, rollback).


czlonkowski:
## Live verification: 10/11 scenarios pass

Run against a real n8n instance (2.28–2.31 range, so its group schema has no `description`). All probe workflows deleted afterwards.

| # | scenario | result |
|---|---|---|
| 1 | create with a group | ✅ group persisted, uuid generated, visible in `full` and `structure` |
| 2 | param edit on a grouped node | ✅ 200, group untouched, no warnings |
| 3 | `removeNode` on a grouped node | ✅ **200** (was 400), group kept on its remaining member + warning naming it |
| 4 | insert a node inside a group | ✅ **200**, nodes/connections intact, only the group dropped, warning quotes n8n's verdict |
| 5 | `setNodeGroups` authoring | ✅ |
| 6 | `setNodeGroups` including the trigger | ✅ fails loudly with n8n's message (`cannot contain trigger nodes`), existing group untouched |
| 7 | `setNodeGroups: []` | ✅ ungroups |
| 8 | group `description` on a pre-2.32 instance | ✅ 200 with description stripped + warning — the ladder's positive control; a non-stripping build 400s here |
| 9 | full update with only `nodes`, then with `[]` | ✅ groups preserved, then ungrouped |
| 10 | rollback across a deletion | ⚠️ **blocked by an unrelated pre-existing bug** — see below |
| 11 | `validate_workflow` with a dangling group member | ✅ warning, `valid: true` |

### Correction to this PR's description

The description lists rollback and version restore among the paths the fix covers. That is accurate at the code level — they share `client.updateWorkflow`, so they inherit the repair and the degradation ladder — but it could **not** be exercised end-to-end through `n8n_workflow_versions`, because `mode: "rollback"` is currently broken independently of this change:

`src/mcp/handlers-n8n-manager.ts` (unchanged by this PR, same line on `main`)
```ts
const client = context ? getN8nApiClient(context) : null;
```

`getN8nApiClient(undefined)` falls through to the env-var path and returns a working client — that is how every other tool resolves it, via `ensureApiConfigured(context)`. The ternary defeats that fallback, so with no `InstanceContext` (the normal stdio / `N8N_API_URL` setup) `client` is always `null`. `rollback` is the only mode that needs it, which is why `list`/`get` work and `rollback` returns `n8n API not configured. Cannot perform rollback without API access.` on any workflow, grouped or not.

Deliberately **not** fixed here to keep this PR scoped; it will get its own PR with a regression test that calls the tool without an instance context.

czlonkowski:
## Rollback fix folded in

The `n8n_workflow_versions` rollback bug flagged in the comment above is now fixed in this PR (`0e8f179`) rather than deferred.

```diff
-const client = context ? getN8nApiClient(context) : null;
+const client = getN8nApiClient(context);
```

The ternary skipped the environment-variable fallback inside `getN8nApiClient` that every other tool reaches through `ensureApiConfigured`. On a plain `N8N_API_URL` setup — the common single-instance case — `client` was always `null`, and since `rollback` is the only mode needing the API it answered `n8n API not configured. Cannot perform rollback without API access.` for every workflow, while `list` and `get` kept working off the local version store.

Multi-tenant isolation is unaffected: it is enforced inside `getN8nApiClient` (which refuses process credentials when `ENABLE_MULTI_TENANT` is set) and by the tenant-scope check at the top of the handler, not by the ternary.

Three regression tests cover it — rollback with no instance context succeeds and passes a real client to the versioning service; a restore failure surfaces its own message rather than a configuration error; and a genuinely unconfigured instance still refuses.

This also unblocks the one live scenario that could not be exercised earlier (rollback across a node deletion, verifying a snapshot whose canvas group predates the deletion is repaired rather than 400ing). It is in the current live test run.

czlonkowski:
## Live verification: 19/19 pass

Re-run against a real n8n instance on the current build, after the earlier pass caught a stale server. Every probe workflow deleted and confirmed gone.

The three that matter most, all previously broken:

| # | scenario | result |
|---|---|---|
| 0 | `setNodeGroups` with `nodeIds: []` alongside a populated `nodeNames` | ✅ creates one group with both nodes, no warnings. On the pre-fix build this returned `success: true`, destroyed the workflow's existing group, created nothing, and warned that the nodes were missing when both were present |
| 16 | group named `Say "hi"`, then made structurally invalid | ✅ ungrouped with the quote intact in the warning, **and the unrelated group survived** — the "destroys every group" path is closed |
| 19 | rollback across a node deletion, with a group | ✅ restores the node and the group's membership; no trace of `n8n API not configured` |

The rest: create with a group (visible in `full` and `structure`), param edit leaves the group untouched, `removeNode` prunes and warns, insert-inside-group ungroups with n8n's verdict quoted, authoring by name, trigger-in-group fails loudly with n8n's message and leaves existing groups untouched, `[]` ungroups, `description` saved with the description stripped and warned about on this pre-2.32 instance, full update preserves groups when `nodeGroups` is omitted and clears them when `[]` is passed, `validate_workflow` warns without invalidating, duplicate member deduped, an authored group with a bogus member fails naming the bad id, `patchNodeField` refuses the node's own `id` while a nested `parameters.…​.id` stays patchable, non-string member and over-long description both rejected locally.

### Unrelated bug noticed during testing

`updateNode` with a bracket path — `updates: { "parameters.assignments.assignments[0].value": … }` — creates a literal sibling key `"assignments[0]"` instead of indexing the array. The numeric-segment form (`.0.`) works correctly. Pre-existing, untouched by this PR, and worth its own issue.

PR #965, opened 2026-07-28
fix: n8n_executions never reads ai_* connection data for AI Agent sub-nodes

JohanDuque:
The execution processor (preview/summary/filtered/full/error modes)
only inspected run.data.main when extracting node input/output data.
Regular nodes connect via `main`, but LangChain AI Agent sub-nodes
(Chat Model, Output Parser, Tool, Memory, Embeddings, etc.) connect to
their parent Agent via special ai_* connection types instead
(ai_languageModel, ai_outputParser, ai_tool, ai_memory, ...). Their
task data lives at run.data.ai_languageModel etc., so every mode
silently returned itemsInput: 0, itemsOutput: 0, data.output: [] for
these nodes regardless of flags.

Also fixes includeInputData, which read a nonexistent `inputData`
field on the run - n8n actually stores this as `inputOverride`, keyed
by connection type the same way `data` is. This made includeInputData
a no-op for every node type, not only AI sub-nodes.

Both files now merge over every connection-type key present (main and
any ai_*) rather than hardcoding `main`, mirroring the same fix
pattern already used elsewhere in this repo for disconnected-node
detection (dynamic iteration over connection types instead of a fixed
list).

Verified against a live n8n Cloud execution with Chat Model, Output
Parser, Tool, and Memory sub-nodes: all four now report correct
itemsInput/itemsOutput/data across preview, summary, filtered, full,
and error modes.

**Update**: added a second fix (commit d227b54) for a related bug found while verifying the first one against live data.

Node-level counts (itemsInput/itemsOutput) already summed across every run of a node, but the code building the actual data.output/data.input arrays only ever read the first run. A node invoked more than once in a single execution — most commonly an AI Agent's Chat Model, called once to decide to call a tool and again to produce the final answer — has one runData entry per invocation, so every invocation after the first was silently dropped from the response while still being counted in the totals. Concretely: itemsOutput: 2 but data.output containing only the first, least useful turn, never the second (the actual answer, sometimes truncated by a max_tokens ceiling).

Fixed by merging across every run of a node (by branch/port index, in run order) instead of reading index 0, in both execution-processor.ts and error-execution-processor.ts. Error detection was extended the same way — an error on a later run is no longer invisible; this surfaced a genuine, previously-hidden error in live testing.

Note on full mode: for nodes invoked multiple times in one execution, full mode responses are now legitimately larger than before, since previously only the first invocation was ever returned (while claiming the correct total in metadata.totalItems). This is the bug being fixed, not new bloat — the size estimate that feeds the tool's preview/recommendation system already accounted for all runs before this fix, so nothing here changes what the tool recommends, only what you get if you explicitly request full anyway. summary/filtered modes remain hard-capped at itemsLimit exactly as before.

### DIFF
--- a/src/services/error-execution-processor.ts
+++ b/src/services/error-execution-processor.ts
@@ -59,11 +59,54 @@ function extractConnectionBranches(connections: unknown): unknown[][] {
 }
 
 /**
- * Get the first output branch's items for a run, across `main` and any
- * `ai_*` connection types.
+ * Whether a node's runData entry has any connection-type data at all
+ * (as opposed to genuinely having zero items in an existing branch).
  */
-function getRunOutputItems(run: any): unknown[] {
-  return extractConnectionBranches(run?.data)[0] || [];
+function nodeHasConnectionData(nodeData: unknown): boolean {
+  if (!Array.isArray(nodeData)) return false;
+  return nodeData.some(run => extractConnectionBranches(run?.data).length > 0);
+}
+
+/**
+ * Get a node's first-branch output items, merged across every run in
+ * run order.
+ *
+ * A node's runData entry is an array of runs, not a single run: nodes
+ * invoked more than once within an execution (e.g. an AI Agent's Chat
+ * Model, called once to decide to call a tool and again to produce the
+ * final answer) get one array entry per invocation. Reading only
+ * `nodeData[0]` silently drops every later invocation's data.
+ */
+function getAllRunsOutputItems(nodeData: unknown): unknown[] {
+  const merged: unknown[] = [];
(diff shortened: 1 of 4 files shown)


czlonkowski:
Thanks @JohanDuque, this was a real gap and your fix is going in. I took it over in #1058 so it could go through the review gauntlet with the release steps: your two commits are carried over with authorship intact, and the follow-up commit addresses what the reviewers found on top (merging keyed by connection type so mixed-type runs stay apart, null ports preserved, a loop instead of push(...items) which overflowed on very large branches, merging bounded by the item limit, inputs truncated too). Closing this one in favour of #1058; you are credited there.

PR #983, opened 2026-08-08
fix(partial-workflow): strip unknown node properties echoed by n8n GET

blouflab:
## Summary

   `n8n_update_partial_workflow` (and full workflow updates) fail with `request/body/nodes/X must NOT have additional properties` when the workflow payload contains node properties that n8n's GET API
   returns but its PUT/PATCH schema rejects.

   ## Problem

   n8n's GET `/workflows/{id}` returns node properties that are server-managed or read-only — e.g. `issues`, `runIndex`, `data`, and even legitimate fields like `onError` and `webhookId` that were missing
   from the local MCP schema. When the MCP tool reconstructs the payload for a partial or full update, these extra properties leak into the request body and n8n rejects it.

## Changes

   - **Added `onError` and `webhookId` to `workflowNodeSchema`** (`src/services/n8n-validation.ts`) so they are recognised as valid instead of being silently dropped.
   - **Introduced `cleanNodeForApi()`** that filters each node through an allow-list of API-safe properties, stripping anything unknown (server-managed echo fields, future additions).
   - **Applied filtering** in both `cleanWorkflowForCreate` and `cleanWorkflowForUpdate`.
   - **Added 4 unit tests** covering the stripping behaviour and the new function.

   ## Verification

   - `npm run typecheck` passes
   - `npm run build` passes
   - `npm test -- tests/unit/services/n8n-validation.test.ts` → 115 tests pass

   Tested with n8n local instance.

### DIFF
--- a/src/services/n8n-validation.ts
+++ b/src/services/n8n-validation.ts
@@ -26,12 +26,59 @@ export const workflowNodeSchema = z.preprocess(normalizeMcpWorkflowNode, z.objec
   notes: z.string().optional(),
   notesInFlow: z.boolean().optional(),
   continueOnFail: z.boolean().optional(),
+  onError: z.enum(['continueRegularOutput', 'continueErrorOutput', 'stopWorkflow']).optional(),
   retryOnFail: z.boolean().optional(),
   maxTries: z.number().optional(),
   waitBetweenTries: z.number().optional(),
   alwaysOutputData: z.boolean().optional(),
   executeOnce: z.boolean().optional(),
+  webhookId: z.string().optional(),
 }));
+
+/**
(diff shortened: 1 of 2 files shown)


czlonkowski:
Thanks @blouflab, this is a real gap: verified live that n8n 2.36.7 rejects any unknown node key on PUT. It is going in via #1059 with your commit and authorship carried over. Two things changed on top: the allowlist is now derived from the zod node schema so the two cannot drift, and a node arm was added to the settings-drift check, which immediately caught customTelemetryTags (accepted by n8n 2.36's node schema, missing from the list). Closing this one in favour of #1059; you are credited there.

PR #996, opened 2026-08-15
chore: upgrade @modelcontextprotocol/sdk to 1.30.0 (v2.69.2)

czlonkowski:
## What

Upgrades `@modelcontextprotocol/sdk` from `1.28.0` to `1.30.0` (releases `v1.29.0` and `1.30.0`). No source changes were required — only the version pins and one stale comment reference.

## Why

Four changes in the two releases reach this server:

1. **SSE keep-alive on `StreamableHTTPServerTransport`.** The transport now writes `: keepalive` comment frames to open SSE streams, so an idle standalone `GET` stream — or a `POST` stream held open through a long-running tool call — is no longer killed by a reverse proxy or a server idle timeout. That failure surfaced to clients as `SSE stream disconnected: TypeError: terminated`. The interval defaults to 15000ms and is configurable via the new `keepAliveMs` option; the default is taken as-is. A companion fix clears the interval on stream cleanup so timers do not accumulate across sessions. This is the default HTTP path (`SingleSessionHTTPServer`), not the deprecated `USE_FIXED_HTTP` one.
2. **`Content-Type` validated by parsed media type** instead of a substring match, so a header that merely contains an accepted type no longer passes.
3. **Zod 3.25 method-literal fix** — relevant because the repo pins `zod` to the exact version `3.25.67`.
4. **Zod validation issues prioritized and formatted** rather than surfaced raw on tool input.

## Behavior change worth knowing

stdio now enforces a **10 MB inbound read-buffer cap** (new SDK default). A single JSON-RPC message larger than that makes the transport emit an error and close instead of buffering without bound. This applies to messages the server *reads*, not to the tool results it writes back, so response size is unaffected; the realistic path to the limit is an oversized `n8n_update_full_workflow` body. It matches the `10mb` JSON body limit the HTTP transport already applies, so both transports now refuse the same size. Configurable via a third `StdioServerTransport` constructor argument, which this server does not set.

## Pins updated in all three places

`package.json`, `package.runtime.json` (the manifest published to npm) and the `Dockerfile` builder stage — so a runtime-only install and a Docker build compile against the same SDK as the tested build. This is the drift class that #339 in the changelog already caught once.

## Not fixed by this

`@hono/node-server` stays at `1.19.14`. The upgrade widens the SDK's own range to `^1.19.9 || ^2.0.5`, but `@n8n/n8n-nodes-langchain@2.32.4` carries its own `@modelcontextprotocol/sdk@1.26.0` with the narrower range, so the resolution does not move. The advisory against `<1.19.15` covers path traversal in `serve-static` on Windows, which is not code this server reaches.

## Verification

- `npm run typecheck` — clean
- `npm run build` — clean
- `npm run test:unit` — 168 files, 5696 passed, 35 skipped, 0 failed
- `vitest run tests/integration/mcp-protocol/` — 7 files, 112 passed, 2 skipped, 0 failed (the pre-upgrade baseline had 3 timeout flakes in this suite; they did not recur)
- `npm run test:cjs-runtime` — passes
- `tests/unit/http-server/method-not-found.test.ts` enumerates every method literal in the SDK's `types.js` and asserts the namespace gate covers them. It passes against 1.30.0, so the comment in `src/http-server-single-session.ts` was updated to name 1.30.0.


czlonkowski:
Follow-up commit d8732987 overrides two of the SDK's new defaults rather than inheriting them.

**stdio read-buffer cap raised to 64 MB.** 1.30.0 introduced a 10 MB cap where stdio was previously unbounded, and the failure mode is a transport close, not a tool error the client can report — the session dies. stdio is the npx channel: no telemetry, and the version is cached on the user's machine, so a workflow body that worked before the upgrade becoming a dead session after it would land on the slowest path to a fix. stdio's peer is the user's own MCP client over a local pipe rather than an untrusted network caller, so a tight bound buys less here than on HTTP. The HTTP transport keeps its `10mb` JSON body limit, which answers an oversized body with an error instead of a disconnect.

**stdio errors are now logged.** Nothing was attached to `onerror`, so a transport-level failure — the new cap among them — ended the session with no diagnostic. Verified against `shared/protocol.js:225-229`: `connect()` captures a handler set beforehand and chains it ahead of its own `_onerror`, so this does not displace the SDK's error propagation.

**SSE keep-alive interval pinned at the call site** (15000ms, the SDK's current default) so a later change to that default cannot move this server's behavior without appearing in a diff.

No new tests. Both changes are configuration values whose names `tsc` already validates, and an assertion that a constructor received a literal would fail on any deliberate change while catching nothing else.

Re-verified after the change: typecheck clean, build clean, 1013 unit tests across the mcp and http-server suites pass, and the MCP protocol integration suite is 112 passed / 2 skipped / 0 failed.

Issue #984, opened 2026-08-09
SSRF address denylist matches hostnames that are not IP addresses


czlonkowski:
`PRIVATE_IP_RANGES` in `src/utils/ssrf-protection.ts` is regex-tested against the raw hostname string in `validateUrlSync`, without gating on `net.isIPv4()`. A DNS label that happens to look like a blocked first octet therefore matches:

```
http://247.example.com   -> blocked: Private IP addresses not allowed
http://224.foo.com       -> blocked: Private IP addresses not allowed
```

This is long-standing behavior — `10.example.com` and `127.example.com` were already refused the same way — but v2.69.0 widened the collision surface from 5 first-octet values to those plus `224`–`255` and the `100.64`–`100.127` two-label prefixes.

Impact is availability-only and it fails safe: such hostnames are refused rather than allowed, and they are not realistic in practice. It was deliberately left out of the v2.69.0 security release because gating on `isIPv4` *loosens* existing behavior, which does not belong in a security patch.

Fix: gate the IPv4 range checks on `isIPv4(hostname)` in the sync validator, mirroring how `isPrivateOrMappedIpv6` already gates on `isIPv6`. The async validator is unaffected — it tests `resolvedIP`, which is always a literal.

Note that this is a deliberate behavior change: hostnames like `10.example.com` would start resolving normally through the sync pre-filter, with the async validator and DNS-pinned transport remaining the real control.

PR #1040, opened 2026-08-30
fix(ssrf): agree on loopback literals under WEBHOOK_SECURITY_MODE=moderate

### LINKED ISSUE #1033: SSRF checks disagree on literal 127.0.0.1 under WEBHOOK_SECURITY_MODE=moderate
Found while extending the \`n8n_test_workflow\` wire tests in PR #1031.

\`validateUrlSync\` (used by \`validateInstanceContext\` for \`x-n8n-url\`) has no loopback allowance under \`moderate\`, so a literal \`http://127.0.0.1:<port>\` falls into the private-range block, while the async \`validateResolvedAddress\` path (used by the official-MCP client's own endpoint check) allows localhost under \`moderate\`. Under the same mode \`http://localhost:<port>\` passes the sync check and \`http://127.0.0.1:<port>\` does not.

The tests work around it with a block-scoped \`WEBHOOK_SECURITY_MODE=permissive\`. The two validators should agree; whichever way the decision goes, the docs for \`WEBHOOK_SECURITY_MODE\` should state it.

maximilliangrand:
## Issue

Fixes #1033.

Under `WEBHOOK_SECURITY_MODE=moderate` the sync and async SSRF validators returned opposite verdicts for the same host. Reproduced against `f895e5e` on main:

```
http://localhost:5678    sync=true    | async=true
http://127.0.0.1:5678    sync=false   | async=true   (sync: "Private IP addresses not allowed")
http://0.0.0.0:5678      sync=false   | async=true
http://[::1]:5678        sync=false   | async=true
```

This also contradicts the documented contract: `docs/SECURITY_HARDENING.md` and `docs/SELF_HOSTING.md` both say `moderate` allows localhost.

## Root cause

Line numbers below are on `f895e5e`.

`SSRFProtection.validateUrlSync` (`src/utils/ssrf-protection.ts:571`) rejects localhost under `strict` at `:613` but has no matching `moderate` allowance, so a loopback literal falls through to the `PRIVATE_IP_RANGES` check at `:621` (which contains `/^127\./` and `/^0\./`) or to `isPrivateOrMappedIpv6` at `:634` for `::1`. The async `validateResolvedAddress` (`:371`) computes `isLocalhost` at `:409` and returns valid for `moderate` at `:423`, ahead of both gates. Only `localhost` slipped through the sync path, because it is the one spelling that is neither an IPv4 literal nor IPv6.

The sync validator is the only gate on `x-n8n-url` inside `validateInstanceContext` (`src/types/instance-context.ts:179`), so the disagreement is user visible.

## Fix

Both validators now share one `isLoopbackHost()` predicate, and `validateUrlSync` gets the `moderate` allowance in the same position the async path uses, ahead of the private-range and IPv6 gates. The loopback prefix test is gated on `isIPv4`, so a DNS name such as `127.example.com` is still handed to the resolving validator rather than allowed outright, matching the existing #984 reasoning.

`::1` is localhost, so it is now allowed under `moderate` exactly as the async path already allowed it. The strict-mode rejection is unchanged and is now asserted on its own rather than bundled with `fe80::`/`fc00::`/`fd00::`.

Two notes on blast radius, since this file is security sensitive:

* The metadata and IPv6 tunneling gates run before any mode branch, so they are untouched. `strict` is unchanged in what it blocks.
* Sharing the predicate makes one small behaviour change on the async side: a hostname whose DNS record is `0.0.0.0` now counts as loopback under `moderate`, where it was previously refused by the private-range gate. A hostname resolving to `127.0.0.1` was already allowed there, and `0.0.0.0` was already in `LOCALHOST_PATTERNS`, so this closes an inconsistency rather than opening a new class. Happy to keep the async side byte-identical instead if you would rather.

Per the issue's second ask, the `WEBHOOK_SECURITY_MODE` row in `docs/SECURITY_HARDENING.md` now spells out which loopback forms `moderate` accepts. The `n8n_test_workflow` wire tests drop the block-scoped `WEBHOOK_SECURITY_MODE=permissive` workaround this defect forced on them and run under `moderate`.

## Tests

New in `tests/unit/utils/ssrf-protection.test.ts`:

* `should agree with validateWebhookUrl on loopback literals in moderate mode` walks `localhost`, `127.0.0.1`, `127.0.0.2`, `0.0.0.0` and `[::1]` through both validators and requires the same verdict.
* `should keep loopback literals blocked in strict mode`.
* `should not treat a hostname merely starting with 127. as loopback`.
* `should reject the IPv6 loopback in strict mode only (it is localhost)`.

Fail then pass, with `src/utils/ssrf-protection.ts` reverted and the tests kept:

```
× should agree with validateWebhookUrl on loopback literals in moderate mode
  -> sync url=http://127.0.0.1:5678 reason=Private IP addresses not allowed
     (use WEBHOOK_SECURITY_MODE=permissive if needed): expected false to be true
× should reject the IPv6 loopback in strict mode only (it is localhost)
  Tests  2 failed | 155 passed (157)
```

The wire tests are a second regression signal: reverting the source with the `moderate` switch kept fails 4 of them.

With the fix applied:

```
npx vitest run tests/unit/utils/ssrf-protection.test.ts
  Tests  157 passed (157)

npx vitest run tests/unit/utils/ssrf-protection.test.ts tests/unit/mcp/handlers-test-workflow.test.ts
  Test Files  2 passed (2)
  Tests  190 passed (190)

npx tsc --noEmit    # clean
```

Every other suite that touches SSRF also passes: `flexible-instance-security`, `multi-tenant-integration`, `instance-context-multi-tenant`, `http-server/ssrf-gate`, `ssrf-pinned-fetch`, the three trigger handlers, `handlers-official-tools`, `official-mcp-access`, `n8n-official-mcp-client`, `n8n-version` and `n8n-api-client`, for 587 passed.

Full unit suite on this machine: `9 failed | 179 passed` test files, with an identical list of failing files on a clean `f895e5e` and on this branch. All 186 failures are `no such module: fts5` or `Cannot find package 'better-sqlite3'`, from this box having no prebuilt `better-sqlite3` binary and falling back to sql.js. Nothing in this change caused them.


czlonkowski:
Thanks @maximilliangrand, this is a real fix and it is going in. I took it over in #1056 so it could run through the review gauntlet with the maintainer's release steps (version bump, changelog): your commit is carried over with authorship intact, the unrelated em-dash comment rewrites were dropped, and one reviewer finding was applied (the resolved-address check keeps its old narrower loopback rule so a DNS name resolving to 0.0.0.0 stays refused in moderate mode). Closing this one in favour of #1056; you are credited there.

PR #1060, opened 2026-09-02
Validate versioned workflows against their own node schemas

### LINKED ISSUE #1037: [Bug] Notion node validator false positive: required property 'title' reported empty when set via expression (regression 2.69.2 → 2.72.0)
## Summary

Since `2.72.0`, workflow validation reports a false positive on the **Notion** node
(`n8n-nodes-base.notion`, `typeVersion: 2.2`, `resource: databasePage`): a required
`title` property that **is** set — via an n8n expression — is reported as empty.

```
<node>: required property 'title' cannot be empty
```

The same workflow validates clean on `2.69.2`.

## Reproduction

Validate this workflow with `validate_workflow`:

```json
{
  "name": "notion-title-expression-repro",
  "nodes": [
    { "parameters": {}, "id": "00000000-1111-2222-3333-444444444444",
      "name": "Manual Trigger", "type": "n8n-nodes-base.manualTrigger",
      "position": [0, 0], "typeVersion": 1 },
    { "parameters": {
        "resource": "databasePage",
        "databaseId": { "__rl": true, "mode": "list",
          "value": "00000000-0000-0000-0000-000000000000",
          "cachedResultName": "example-db" },
        "propertiesUi": { "propertyValues": [
          { "key": "URL|url",         "urlValue":    "={{ 'https://example.com' + $json.path }}" },
          { "key": "title|title",     "title":       "={{ $json.title }}" },
          { "key": "abstract|rich_text", "textContent": "={{ $json.abstract }}" }
        ] },
        "options": {} },
      "id": "11111111-2222-3333-4444-555555555555",
      "name": "Store Page", "type": "n8n-nodes-base.notion",
      "position": [220, 0], "typeVersion": 2.2 }
  ],
  "connections": { "Manual Trigger": { "main": [[{ "node": "Store Page", "type": "main", "index": 0 }]] } }
}
```

**Expected** (and what `2.69.2` returns): `valid: true`, `errorCount: 0`.
**Actual** (`2.72.0` and later): one error, `required property 'title' cannot be empty`.

## Which versions

An automated nightly gate validates our whole workflow set against each new image
before promoting it. Verdicts for the affected workflow:

| Version | Verdict |
|---|---|
| 2.68.4, 2.69.0, **2.69.2** | green |
| **2.72.0**, 2.73.0, 2.74.1, 2.75.1 | red — this error |

So the regression lands between `2.69.2` and `2.72.0` (we never tested 2.70.x/2.71.x).
Nothing changed on our side: the same stored workflow flips verdict with the image alone.

## What narrows it down

In the **same node**, two other property types also hold expressions and are **not**
flagged — only `title` is:

| Property | Value field | Flagged? |
|---|---|---|
| `URL\|url` | `urlValue` | no |
| `abstract\|rich_text` | `textContent` | no |
| `title\|title` | `title` | **yes** |

So this does not look like a blanket rejection of expression values, but something
specific to how the required `title` property is read. Note that each Notion property
type stores its value under a different key (`urlValue`, `textContent`, `title`, `date`…).

Minor, possibly related: the error message lower-cases the node name
(`store abstract notion` for a node actually named `Store Abstract Notion`).

## Impact

We hold at `2.69.2` because our promotion gate treats a new validation error as a
regression, and we would rather not promote a validator that reports a healthy
workflow as broken. Happy to run any patch or debug build against the same set.

justadityaraj:
## Context

Notion 3 made the database-page `title` field required. Workflows saved with Notion 2.2 were still being validated against that latest schema, so a valid nested title expression produced `Required property 'Title' cannot be empty`.

Review also exposed a related path: persisted `*Tool` variants resolve version rows through their base node, whose historical schema does not contain the generated `toolDescription` property.

Fixes https://github.com/czlonkowski/n8n-mcp/issues/1037

## Changes

- resolve versioned node properties from the workflow node's `typeVersion`
- preserve the generated `toolDescription` property when a Tool variant uses its base node's historical schema
- fall back to the latest schema when an exact version row is unavailable
- cover the Notion 2.2, Google Calendar Tool 1.3 strict-profile, and missing-version paths

## User impact

Older versioned nodes are validated against the schema they actually execute with, including stored Tool variants, while current nodes retain their existing validation.

## Verification

- Tool-variant regression test failed before the fix with `Property 'toolDescription' won't be used - not visible with current settings`
- Local validator/repository suite: 374 tests passed
- Local `npm run typecheck`, `npm run build`, and `git diff upstream/main...HEAD --check`: passed
- GitHub Actions: unit coverage, offline integration tests, lint, typecheck, CommonJS runtime, secret scan, and Codecov patch coverage passed
- Docker publish jobs stopped at registry login because fork PRs do not receive the repository credentials

### DIFF
--- a/src/services/workflow-validator.ts
+++ b/src/services/workflow-validator.ts
@@ -653,10 +653,16 @@ export class WorkflowValidator {
           '@version': node.typeVersion || 1,
           ...node.parameters
         };
+        const versionProperties = nodeInfo.isVersioned && node.typeVersion != null
+          ? this.nodeRepository.getNodeVersion(normalizedType, String(node.typeVersion))?.propertiesSchema
+          : null;
+        const properties = versionProperties && nodeInfo.isToolVariant
+          ? [...nodeInfo.properties.filter((property: any) => property.name === 'toolDescription'), ...versionProperties]
+          : versionProperties || nodeInfo.properties;
         const nodeValidation = this.nodeValidator.validateWithMode(
           node.type,
           paramsWithVersion,
-          nodeInfo.properties || [],
+          properties || [],
           'operation',
           profile as any
         );
--- a/tests/unit/services/workflow-validator-version-schema.test.ts
+++ b/tests/unit/services/workflow-validator-version-schema.test.ts
@@ -1,0 +1,143 @@
+import { describe, expect, it, vi } from 'vitest';
+import { EnhancedConfigValidator } from '@/services/enhanced-config-validator';
+import { WorkflowValidator } from '@/services/workflow-validator';
+import type { NodeRepository } from '@/database/node-repository';
+
+vi.mock('@/utils/logger');
+
+describe('WorkflowValidator version-specific schemas', () => {
+  it('validates a Notion 2.2 database page against its own schema', async () => {
+    const title = {
+      displayName: 'Title',
+      name: 'title',
+      type: 'string',
+      default: '',
+      required: true,
+      displayOptions: { show: { resource: ['databasePage'], operation: ['create'] } },
+    };
+    const properties = [
+      { displayName: 'Resource', name: 'resource', type: 'options', default: 'page' },
+      { displayName: 'Operation', name: 'operation', type: 'options', default: 'create', displayOptions: { show: { resource: ['databasePage'] } } },
+      title,
+    ];
+    const repository = {
+      getAllNodes: vi.fn(() => []),
+      getNode: vi.fn((type: string) => type === 'nodes-base.manualTrigger'
+        ? { nodeType: type, displayName: 'Manual Trigger', package: 'n8n-nodes-base', isTrigger: true, isVersioned: false, outputs: ['main'], properties: [] }
+        : { nodeType: type, displayName: 'Notion', package: 'n8n-nodes-base', version: 3, isVersioned: true, outputs: ['main'], properties }),
+      getNodeVersion: vi.fn(() => ({ propertiesSchema: properties.map(property => property === title ? { ...property, required: false } : property) })),
+    } as unknown as NodeRepository;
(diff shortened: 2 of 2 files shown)


czlonkowski:
Thanks for the focused fix. The Notion 2.2 path is valid, but I found one regression that should be addressed before merge.

`src/services/workflow-validator.ts:656-658` applies `getNodeVersion()` to every versioned node. For persisted `*Tool` variants, the repository intentionally resolves version rows to the base node (`src/database/node-repository.ts:902-912`), while the variant schema adds `toolDescription` (`src/services/tool-variant-generator.ts:67-89`). The selected historical base schema therefore drops a valid variant-only property, and strict validation can emit:

`Property 'toolDescription' won't be used - not visible with current settings`

I reproduced this with `n8n-nodes-base.googleCalendarTool` at `typeVersion: 1.3`; the bundled database currently has 75 versioned stored Tool variants following the same base-version mapping.

Please preserve or merge variant-only properties when selecting the historical base schema (or handle `nodeInfo.isToolVariant` separately), and add a strict-profile regression test with a configured `toolDescription`. A missing-exact-version fallback test would also cover the currently untested branch.

One process item as well: `AGENTS.md` requires the exact attribution `Conceived by Romuald Członkowski - https://aiadvisors.pl/en` in both the PR description and commit message; both currently use `www.aiadvisors.pl/en`.

Local verification: 317 relevant validator/database tests and typecheck passed.

justadityaraj:
Fixed in `212082f3`:

- historical base schemas now retain the stored Tool variant's generated `toolDescription`
- added the strict `googleCalendarTool` 1.3 regression and missing-exact-version fallback coverage
- amended the commit and PR description to use the required attribution URL

Verified: 374 relevant tests, typecheck, build, and diff check passed. The full unit run still hits the existing Windows FTS5/symlink/shell limitations and does not exit cleanly.

PR #1061, opened 2026-09-02
Keep catalog reads scoped to the request instance

### LINKED ISSUE #1034: Remaining env-fallback reads on a url + MCP-token context without x-n8n-key
Follow-up from the PR #1031 reviews. PR #1031 made every Public-API-dependent path of \`n8n_test_workflow\`, the consent write, and the project lookup behind \`n8n_list_catalog\` / the data-table column actions refuse (or reroute to the MCP server) when a per-request context carries \`x-n8n-url\` and \`x-n8n-mcp-token\` but no \`x-n8n-key\`, instead of falling back to the environment's Public API client. Two residuals remain, both read-only:

1. \`handleListCatalog\` with \`kind: 'tags'\` still reads tags through the env-fallback client on such a context (\`src/mcp/handlers-official-tools.ts\`, tags branch).
2. \`handleListCatalog\` short-circuits on its own \`getN8nApiClient\` null check and reports the generic "n8n API not configured" text rather than \`PUBLIC_API_CONTEXT_HINT\`, so the two paths word the same condition differently.

Suggested fix: apply \`publicApiMatchesContext\` in \`handleListCatalog\` before any Public API read (refuse with the shared hint, or route tags through the MCP server's \`list_workflow_tags\`), and consider making \`getN8nApiClient\` return null for a url-without-key context so the rule holds for every management tool.

justadityaraj:
## Problem / Context

A request context containing an n8n URL and MCP token but no Public API key is authoritative for the official MCP client. The catalog handler still resolved the Public API client before branching, so tag reads could fall back to the operator's environment instance. The same early lookup replaced the shared same-instance hint with a generic configuration error for project requests without an environment client.

Fixes https://github.com/czlonkowski/n8n-mcp/issues/1034

## Fix / Changes

- Refuse tag reads with the shared context hint before resolving any Public API client.
- Resolve the Public API client only inside the tags branch.
- Let project requests continue through the existing project resolver and its official-MCP fallback.
- Cover both residual context shapes with handler regressions.

## User impact

Per-request catalog calls no longer read tags from a different environment-configured n8n instance, and project failures consistently explain that `x-n8n-key` must accompany `x-n8n-url`.

## Verification

- New focused regressions failed on current upstream `main`: the environment API getter was called for tags, and projects returned the generic setup error.
- `npx vitest run tests/unit/mcp/handlers-official-tools.test.ts --reporter=dot` — 39 passed.
- Relevant handler, exposure, and official-MCP integration selection — 66 passed, 2 environment-gated tests skipped.
- `npm run typecheck` — passed.
- `npm run build` — passed.
- `npm run test:unit` executed the full unit set but did not exit on this Windows host after emitting all files; unrelated existing failures were confined to unavailable SQLite FTS5, POSIX shell/path assumptions, and Windows symlink privileges.

### DIFF
--- a/src/mcp/handlers-official-tools.ts
+++ b/src/mcp/handlers-official-tools.ts
@@ -170,10 +170,13 @@ export async function handleListCatalog(args: unknown, context?: InstanceContext
     return { success: false, code: 'INVALID_ARGS', error: parsed.error.issues.map(i => `${i.path.join('.') || 'input'}: ${i.message}`).join('; ') };
   }
   const { kind, query, limit } = parsed.data;
-  const api = getN8nApiClient(context);
-  if (!api) return { success: false, code: 'NOT_CONFIGURED', error: 'n8n API not configured. Set N8N_API_URL and N8N_API_KEY.' };
 
   if (kind === 'tags') {
+    if (!publicApiMatchesContext(context)) {
+      return { success: false, kind, code: 'NOT_CONFIGURED', error: PUBLIC_API_CONTEXT_HINT } as McpToolResponse;
+    }
+    const api = getN8nApiClient(context);
+    if (!api) return { success: false, kind, code: 'NOT_CONFIGURED', error: 'n8n API not configured. Set N8N_API_URL and N8N_API_KEY.' } as McpToolResponse;
     try {
       const tags = (await api.listTags({ limit: 250 })).data.map(t => ({ id: String(t.id), name: t.name }));
       return { success: true, kind, backend: 'public-api', data: { items: filterItems(tags, query, limit) } } as McpToolResponse;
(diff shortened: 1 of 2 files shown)


czlonkowski:
Reviewed the current `fc0eba1` revision. The code change looks correct and I found no blocking correctness, security, or privacy issue.

The new guard in `src/mcp/handlers-official-tools.ts:175-179` runs before `getN8nApiClient()`, so a URL + MCP-token context without `x-n8n-key` cannot read tags through the environment API. The projects path continues through the existing context-aware resolver. The added regressions cover both residual cases from #1034.

The only requested change is metadata/process: `AGENTS.md` requires the exact attribution `Conceived by Romuald Członkowski - https://aiadvisors.pl/en` in both the PR description and commit message; both currently use `www.aiadvisors.pl/en`. Please update the PR description and commit attribution, unless the maintainers explicitly decide that the final squash commit is sufficient for the commit-message requirement.

Local verification: 58 relevant unit tests and typecheck passed. The socket-bound wire tests were not rerun locally because the review sandbox cannot host their fake server; the full GitHub test workflow passed. Once the attribution is resolved, I am comfortable with this PR being merged.

PR #1073, opened 2026-09-05
fix: validate workflow node shapes before graph traversal (v2.84.1)

### LINKED ISSUE #1071: n8n_create_workflow: unhandled TypeError (toLowerCase / startsWith) on malformed nodes[] entries instead of a validation error
## Summary

`n8n_create_workflow` returns an unhandled `TypeError` instead of a validation error when the
`nodes` array contains an entry that is not a well-formed node object **and** that entry is
referenced from `connections`.

The error surfaces to the client as:

```
Cannot read properties of undefined (reading 'toLowerCase')
```

A related shape returns `node.type?.startsWith is not a function`.

Both look like connection/type resolution running before (or independently of) the node-shape
validation that already exists and works correctly for the same inputs.

## Reproduction

n8n-mcp `2.79.0`, node `v22.23.2`, HTTP transport.

**1. Crash — string entry in `nodes[]` that `connections` references:**

```json
{
  "name": "repro",
  "nodes": [
    {"id": "1", "name": "Manual Trigger", "type": "n8n-nodes-base.manualTrigger",
     "typeVersion": 1, "position": [0, 0], "parameters": {}},
    "strayString"
  ],
  "connections": {
    "Manual Trigger": {"main": [[{"node": "strayString", "type": "main", "index": 0}]]}
  }
}
```

→ `Cannot read properties of undefined (reading 'toLowerCase')`

**2. Crash — node whose `type` is not a string:**

```json
{
  "name": "repro",
  "nodes": [
    {"id": "1", "name": "Manual Trigger", "type": "n8n-nodes-base.manualTrigger",
     "typeVersion": 1, "position": [0, 0], "parameters": {}},
    {"id": "2", "name": "Num Type", "type": 123, "typeVersion": 1,
     "position": [200, 0], "parameters": {}}
  ],
  "connections": {}
}
```

→ `node.type?.startsWith is not a function`

## What already works (for contrast)

These same malformed inputs are handled correctly when the bad entry is *not* wired up, which is
why this reads as a missing guard rather than a missing validator:

| Input | Result |
|---|---|
| String entry in `nodes[]`, not referenced by `connections` | `Workflow validation failed` ✅ |
| Node object missing `type`, not referenced | `Workflow validation failed` ✅ |
| Connection pointing at a node name that does not exist (all nodes valid) | `Workflow validation failed` ✅ |
| Connection key naming a node that does not exist | `Workflow validation failed` ✅ |
| **String entry in `nodes[]` that `connections` references** | **TypeError** ❌ |
| **Node with non-string `type`** | **TypeError** ❌ |

`validate_workflow` handles every one of the above cleanly (`valid: false` with 2 errors), so the
gap is specific to the create path.

## Why it matters

Any client that builds workflow JSON programmatically can emit a malformed `nodes` entry — a
partially recovered JSON payload, a template that interpolated wrong, a serialisation bug. The
current behaviour turns that into an opaque runtime error that looks like a server fault, so
clients cannot distinguish "my payload was malformed" from "the server is broken", and it is not
actionable without server logs.

## Suggested fix

Before connection/type resolution in the create path, reject entries in `nodes` that are not
objects with string `name` and `type`, and surface them through the same
`Workflow validation failed` response used elsewhere. Optionally guard the accessors
(`node?.type?.toLowerCase?.()`) as defence in depth.

maximilliangrand:
## Summary

Fixes #1071. A string or null entry in `nodes`, or a node with a non-string `type`, can reach graph traversal or short-form type detection before validation and return a TypeError instead of an actionable workflow error.

Validate node shapes with the existing node schema before inspecting connectivity. Collect indexed errors and return them through the existing `Workflow validation failed` response. Graph checks use normalized copies, preserving accepted serialized inputs without mutating the caller's workflow. Guard the create handler's short-form detection while preserving its existing full-form guidance.

Regression tests cover malformed entries with and without connections, multiple invalid entries, normalized inputs, and valid workflow creation. Handler tests use the real structure validator and verify that malformed input never calls the create API.

## Validation

- 19 new service cases fail before the service fix; eight handler cases reproduce the original TypeErrors before the combined fix.
- `npm run typecheck` and `npm run build`: passed on Node 22.
- `npm run test:coverage -- tests/unit`: 6,318 passed, 35 skipped; 84.54% line/statement, 86.33% branch, and 84.9% function coverage, above the configured thresholds.
- Offline integration suite using the same exclusions as fork CI: 511 passed, 47 skipped. Live n8n/AI suites were excluded; Docker and optional lifecycle suites reported skips.

GitHub CI note: the secret scan and CommonJS runtime check pass. The two Docker publication workflows fail at the Docker Hub login step (`Username and password required`) on this fork PR, before building the changed code.

AI disclosure: investigation, implementation, regression tests, and this PR description were primarily written with Codex.


czlonkowski:
Thanks for this — the diagnosis is right and the shape of the fix is the right one: parsing each node before the graph checks is what makes the `node.type.startsWith` dereferences downstream provably safe, rather than scattering `typeof` guards through the traversal.

I verified it against a 14-case matrix on `main` vs this branch: `main` threw a TypeError in 5 of them, this branch in none, and valid workflows — including n8n's GET-echoed extra node properties and JSON-string-serialized `parameters`/`position` — still validate unchanged.

I've pushed four commits to the branch (maintainer edits were enabled) and run the review gauntlet over the result:

- **The error is now one line instead of a serialized Zod issue array.** `Invalid node at index 1: "type": Expected string, received number` rather than ~15 lines of JSON. Since the shape check returns early, that string is the entire response an MCP client sees, so it was worth condensing. Your `toContain('"type"')` assertions still hold.
- **A `null` entry still threw in `n8n_update_full_workflow`.** The credential-preservation merge reads `node.credentials` off every submitted entry ~70 lines before the validator runs, so `[validNode, null]` produced `Cannot read properties of null (reading 'credentials')` — the same class, and the shape most likely to come out of a dropped array element. Guarded, with tests.
- **Test hardening.** The exact-message assertion pinned Zod 3 wording (the MCP SDK also accepts Zod 4, which words it differently), and the no-mutation fixture used only schema-declared keys, so it could not have caught a future refactor that strips the caller's nodes. Both fixed.
- **Merged `main` and bumped to v2.84.1** with a changelog entry crediting you.

The early return — a malformed node suppressing the other structural errors — I kept as you wrote it. Connectivity computed over a node of unknown shape is misleading, and the accept/reject outcome is unchanged either way.

One thing your fix surfaced: the same defect class was still open in `n8n_update_partial_workflow` and `validate_workflow` (the latter contrary to what #1071 assumed). That's #1092, fixed in #1093, which merges after this one.

czlonkowski:
Live-verified against a real n8n instance (n8n 2.38.5) with this branch built and served through the MCP server, not just unit tests:

| case | response |
|---|---|
| create, string entry, no connections | `Invalid node at index 1: Expected object, received string` |
| create, string entry, referenced by connections | same — this is the shape that used to throw `Cannot read properties of undefined (reading 'toLowerCase')` |
| create, `null` entry, either way | `Invalid node at index 1: Expected object, received null` |
| create, `type: 123`, either way | `Invalid node at index 1: "type": Expected string, received number` |
| full update, `nodes: [valid, null]` | `Invalid node at index 1: Expected object, received null` — previously `Cannot read properties of null (reading 'credentials')` |
| full update, valid follow-up | applied normally |

No workflow was created for any of the six malformed create attempts (confirmed against the workflow list), and an ordinary create → validate → partial update → fetch → delete lifecycle on the same instance was unaffected.

Issue #1094, opened 2026-09-11
Malformed connection shapes still surface as internal TypeErrors in validate_workflow and the workflow validators


czlonkowski:
## Summary

#1071 / #1092 closed the malformed-**node** case: an entry in `nodes` that is not a well-formed
node object is now reported as a validation error rather than a TypeError. The same class is
still open one level down, in **connections** — the shapes are read without being checked, so a
malformed connection object produces an internal dereference message instead of an actionable
error.

Found by the review gauntlet on #1093. All of the following were executed against
`fix/diff-engine-malformed-add-node` (which already contains the #1092 fixes).

## Reproductions

Workflow: two valid nodes, `Start` (manual trigger) and `B` (set).

`validateWorkflowStructure` (`src/services/n8n-validation.ts`) — the `n8n_create_workflow` and
`n8n_update_full_workflow` path:

```
connections: { Start: null }   →  throws  Cannot convert undefined or null to object
```

The connection traversal runs before `validateWorkflowConnections` parses the shape, so the
throw escapes the validator entirely rather than becoming an indexed error.

`WorkflowValidator.validateWorkflow` (`src/services/workflow-validator.ts`) — the
`validate_workflow` tool:

```
connections: { Start: null }          →  Workflow validation failed: Cannot convert undefined or null to object
connections: { Start: { main: [{}] } } →  Workflow validation failed: outputConnections.forEach is not a function
```

As with the node case, the single try/catch around the run means this is the **only** error the
caller sees — every other check is skipped.

## Also reported by the review, not re-verified here

- `validateWorkflowStructure`: a Switch node with one rule and `connections.Switch.main: [null]`
  reaching `Cannot read properties of null (reading 'length')`.
- `workflow-validator.ts`: `{ Start: { main: [[null]] } }`, and the same class reachable with
  `validateConnections: false`, because expression checking still calls `nodeHasInput`.
- `src/services/node-sanitizer.ts`: a Switch v3.2 node with `parameters.rules.rules: [null]`
  throwing on both the diff path (`Failed to apply operation: Cannot read properties of null
  (reading 'conditions')`) and the create/full-update structure validation.
- `validate_workflow`'s node-shape gate accepts a *present* non-string `name` (e.g.
  `{toString: null}`), which then throws at connection indexing. Allowing a draft node to omit
  its name need not mean accepting a non-string one.

## Design note from the same review

[...]

Issue #1092, opened 2026-09-11
n8n_update_partial_workflow: addNode with a malformed node payload reports a TypeError instead of an operation error


czlonkowski:
## Summary

`n8n_update_partial_workflow` answers a malformed `addNode` payload with an internal `TypeError`
wrapped as a diff-engine error, against operation index `-1`. This is the same defect class as
#1071 (fixed for `n8n_create_workflow` in #1073), in the tool that receives the most writes.

## Reproduction

Against a workflow with any nodes, each of these operations:

```json
{ "type": "addNode", "node": "strayString" }
{ "type": "addNode", "node": { "id": "2", "name": "Bad", "type": 123, "typeVersion": 1, "position": [200, 0], "parameters": {} } }
{ "type": "addNode", "node": { "id": "2", "name": "Bad", "typeVersion": 1, "position": [200, 0], "parameters": {} } }
```

returns:

```
{"operation": -1, "message": "Diff engine error: Cannot read properties of undefined (reading 'trim')"}
{"operation": -1, "message": "Diff engine error: node.type.includes is not a function"}
{"operation": -1, "message": "Diff engine error: Cannot read properties of undefined (reading 'includes')"}
```

## Root cause

The diff request schema types an operation's `node` as `z.any()`
(`src/mcp/handlers-workflow-diff.ts`), which it has to: the `addNode` contract is deliberately
looser than n8n's node schema, because `applyAddNode` fills in `id`, `typeVersion` and
`parameters`. So an arbitrary value reaches `WorkflowDiffEngine.validateAddNode`
(`src/services/workflow-diff-engine.ts`), which calls `this.normalizeNodeName(node.name)` and
`node.type.includes('.')` without checking either field first.

## Why it matters

The operation index is lost (`-1`), so a caller applying a batch of operations cannot tell which
one was malformed, and the message describes an internal dereference rather than the payload
problem. A client that builds diff operations programmatically — a partially recovered JSON
payload, a template that interpolated wrong — cannot distinguish its own bad payload from a
server fault, which is exactly what #1071 described for the create path.

## Expected

An operation error naming the field and the offending operation index, e.g.
`addNode requires a string "type" on the node, received a number`.

PR #1099, opened 2026-09-12
fix: accept a null output branch and validate Switch rules.values (v2.84.4)

### LINKED ISSUE #1096: validate_workflow accepts a nullish output branch that n8n_create_workflow rejects
Found by the review gauntlet on #1095. Predates that PR — it is a divergence between the two validators, not a regression.

## The divergence

n8n's own type says a null output branch is legal (`node_modules/n8n-workflow/dist/esm/interfaces.d.ts:313`):

```ts
export type NodeInputConnections = Array<IConnection[] | null>;
```

`validate_workflow` follows that type and accepts it. The write path does not:

```
connections: { Start: { main: [[{ node: "B", type: "main", index: 0 }], null] } }

validate_workflow      → valid: true, no errors
n8n_create_workflow    → Invalid connections: "Start.main.1": Expected array, received null
```

The write-path rejection comes from `connectionArraySchema` in `src/services/n8n-validation.ts`, which is `z.array(z.array(z.object({node, type, index})))` — no `nullable()` on the inner array.

An agent that validates a draft, gets `valid: true`, and then fails to create it has no way to tell which validator is wrong.

## Which one is wrong

Unverified. n8n's TypeScript type permits the null, but the Public API's own schema may not — the two are not the same contract, and n8n-mcp's write schema mirrors what the API accepts, not what the internal type declares. Before loosening anything, someone should POST a workflow with a null branch to a live n8n instance and record the answer.

- If n8n accepts it: `connectionArraySchema` should become `z.array(z.array(...).nullable())` and the divergence closes in favour of the type.
- If n8n rejects it: `validate_workflow` should report it, and the comment in `collectMalformedConnectionErrors` citing `NodeInputConnections` needs correcting — it currently uses that type as the reason for accepting.

Note that #1095 deliberately kept `validate_workflow` accepting this shape, because an existing test pins the tolerance and nothing dereferences it. That decision is only safe while the write path's answer stays visible, which it now is: the create-path message is one readable line rather than a serialized Zod blob.

## While we are in there

`WorkflowValidator.validateWorkflowConnections` at `src/mcp/server.ts:4666` is a private method with no call site. Dead, so it has no user impact, but if it were ever wired up its `e.message.includes('connection')` filter is case-sensitive and would drop three of the five connection shape messages #1095 added — returning `valid: false` with an empty `errors` array. Worth deleting rather than repairing. `src/services/workflow-diff-engine.ts:39` also imports `validateWorkflowConnections` without using it.

### LINKED ISSUE #1097: Switch condition operators are never validated for the shape n8n actually uses (rules.values)
Found while live-testing #1095. Pre-existing; not caused by that PR, and not a crash — a coverage gap.

## The gap

`validateConditionNodeStructure` (`src/services/n8n-validation.ts`) validates Switch v3.2+ condition operators by reading `parameters.rules.rules`:

```ts
const rules = node.parameters?.rules as any;
if (Array.isArray(rules?.rules)) { /* validate each rule's conditions */ }
```

But n8n's Switch node defines `rules` as a `fixedCollection` whose only option is named **`values`** (confirmed against `nodes-base.switch` in the bundled database), so real workflows carry `parameters.rules.values`. Measured across the 2,352 bundled templates:

```
Switch nodes: 448 | rules.values: 319 | rules.rules: 5 | neither: 124
typeVersions: 3.2 → 332, 3.4 → 47, 3.3 → 40, 3 → 20, rest → 9
```

So the operator-structure validation this function performs never runs for the shape 319 of 448 real Switch nodes use. `EnhancedConfigValidator.validateSwitchNodeStructure` does read `rules.values`, but it only checks for the presence of `conditions` and `outputKey` and emits **warnings**, not the operator-type/operation validation `validateFilterConditionOperators` does for IF nodes.

The same asymmetry shows up for malformed entries. After #1095:

```
parameters.rules.rules:  [null]  → error:   "rules.rules[0]: rule is missing or not an object"
parameters.rules.values: [null]  → warning: "Switch rule 1 is missing \"conditions\" property", valid: true
```

Live-verified through the MCP server: a workflow with `rules.values: [null]` validates clean and `n8n_create_workflow` accepts it.

## Why this wasn't fixed in #1095

That PR's scope was payload shapes surfacing as internal TypeErrors. Both paths throw no longer, which closes that defect. Extending operator validation to `rules.values` would newly report on a large share of real Switch nodes, so it needs its own false-positive pass against the template corpus first — the same discipline the July validator FP audit established (falsely-invalid templates 77% → 39%).

## Worth checking while fixing

The 5 templates carrying `rules.rules` are probably broken at runtime, since n8n reads `rules.values` — they would execute with no routing rules at all. If that is confirmed, `validate_workflow` should arguably report a Switch whose rules sit under the wrong key, which would be a genuinely useful new check rather than a stricter one.

czlonkowski:
Closes #1096. Closes #1097.

Both were opened by the review gauntlet on #1095 and left there because each needed evidence that PR did not have. This one has it.

## #1096 — a null output branch is legal n8n data

The two validators disagreed:

```
connections: { Start: { main: [[{ node: "B", type: "main", index: 0 }], null] } }

validate_workflow    → valid: true, no errors
n8n_create_workflow  → Invalid connections: "Start.main.1": Expected array, received null
```

The issue asked for a live POST to settle it. Done, against a real n8n instance:

```
POST /api/v1/workflows  → HTTP 200
GET  /api/v1/workflows/<id>
  → {"Start": {"main": [[{"node": "B", "type": "main", "index": 0}], null]}}
```

n8n accepts it and returns it unchanged, matching its own `NodeInputConnections = Array<IConnection[] | null>`. So `validate_workflow` was right and the write path's `connectionArraySchema` was wrong.

Accepting the shape is the small half. A workflow read back out of n8n could always carry a null branch, so these walks were already reachable before this change:

| site | what it did |
|---|---|
| `validateWorkflowStructure` Switch branch counts | `branch.length` on null |
| diff engine `cleanStaleConnections` | `for (const conn of conns)` on null |
| diff engine `applyRemoveConnection` | `conns.filter(...)` on null |
| diff engine `validateRemoveConnection` | `conns.some(...)` on null |
| diff engine `applyRemoveNode` | `connections.filter(...)`, then `.length` on null |
| diff engine `validateRemoveNode` | `connections.some(...)` on null |
| diff engine node-rename pass | `connectionsAtIndex.length` on null |

That is the defect class #1071 / #1092 / #1094 closed one level up. Widening `WorkflowConnection` in `src/types/n8n-api.ts` is what surfaced the last five — they were invisible while the type claimed branches were never null.

Semantics chosen: a null branch counts as an **unconnected output** wherever branches are counted; an **intermediate** one is preserved so positional indices do not shift; a **trailing** one is trimmed alongside the trailing empty arrays already trimmed there. `replaceConnections` accepts one for the same reason a create does — a caller may send back a shape it read from n8n — and every other non-array branch is still rejected.

Two dead private methods on the MCP server go with it, as the issue suggested: `validateWorkflowConnections` and `validateWorkflowExpressions` are leftovers from tools consolidated into `validate_workflow`, neither had a call site, and the connection one filtered its findings with a case-sensitive `includes('connection')` that would have dropped three of the five connection shape messages #1095 added.

## #1097 — Switch rules live under `values`, not `rules`

The `rules` fixedCollection declares one option and it is named `values`. Measured across the 2,352 bundled templates:

```
Switch nodes: 448 | rules.values: 319 | rules.rules: 5 | neither: 124
```

All 5 `rules.rules` nodes are **typeVersion 1**, where that key with `{value2, operation, output}` entries is the correct legacy shape — the existing `typeVersion >= 3.2` gate already excludes them. The issue's guess that they are broken at runtime was wrong, and a test now pins that they stay unreported.

So the operator-structure validation IF nodes receive never ran on a v3.2+ Switch. An operator missing `type` or `operation`, or naming an operation such as `equals` in its `type` field, passed unreported, and `rules.values: [null]` validated clean and was accepted by `n8n_create_workflow` while the identical entry under `rules.rules` was reported. One `validateSwitchRuleCollection` helper now walks both keys; the `rules.rules` messages are unchanged.

### False-positive pass

Extending a validator to a shape 319 real nodes use needs the corpus check the July audit established. Simulating the new check over every bundled template:

```
944 rules validated across 319 Switch nodes → 0 new errors
(IF-node baseline for comparison: 1 error / 1,665 nodes)
```

**The branch-count check beside it reads the same wrong key and is deliberately left alone**, because the same measurement gives the opposite answer — 51 templates would newly report, and close to all of them falsely:

- 35 carry a `fallbackOutput`, whose extra output branch the check does not model (33 of them exactly one branch over)
- the other 15 have trailing unconnected outputs that n8n simply omits from `connections`

That needs its own fix and is tracked separately.

## Verification

- 7,352 tests pass (one known-flaky timing assertion in `tests/integration/database/performance.test.ts` passes on rerun)
- The 7 new `rules.values` tests and 4 of the 5 new diff-engine tests were confirmed to **fail** against the pre-fix code; the fifth is a guard that non-array branches are still rejected, and correctly passes either way
- `npm run typecheck` and `npm run build` clean


czlonkowski:
Copilot's five points, each acted on or answered with evidence.

**Fixed**

1. `validateOperator` in `node-sanitizer.ts` excluded `any` while `validateOperatorStructure` accepted it — fixed in `0b2f0fc3`, then made structural in `01d783ed`: both now read `FILTER_OPERATOR_TYPES`, so the two lists cannot drift again. Codex found the same defect independently and reproduced it.
2. `ReplaceConnectionsOperation` still required every branch to be an array, making the payload the runtime accepts inexpressible in TypeScript. Widened, and the test now pins it with `satisfies` rather than casting past it — reverting the type turns that test into a compile error.
3. The missing-`type` diagnostic listed six data types while the check accepted seven. Both messages now read from the same constant.

**Fixed, slightly differently than suggested**

4. "Duplicate `rules.values` errors." There is no duplicate error — the second report was a *warning* from `EnhancedConfigValidator`, which described `rules.values: [null]` as "a rule missing its `conditions` property". The condition validator already reports that entry as `rules.values[0]: rule is missing or not an object`, which is what it is; the warning pointed at a repair that would not have helped. Non-rule entries are now skipped in the warning pass, and a real rule missing its conditions still warns.

**Not a defect**

5. "`validateRewireConnection` dereferences `connections[sourceIndex].some(...)` without a guard, so rewiring on a null output slot throws." It does not throw — `if (!connections[sourceIndex])` two lines above catches null, and `applyRewireConnection` reads its slots through `?? []`. Verified against the built code with `main: [null, [{node: "A", ...}]]`:

```
sourceIndex 0 → success: false | "No connections found from "S" on output "main" at index 0"
sourceIndex 1 → success: true
```

A clean operation error and a working rewire, which is the intended behaviour.

**Real, and filed rather than folded in**

6. "A null-only output map passes the disconnected-node scan, because the scan adds every connection source to `connectedNodes` before finding an edge." Correct, and Codex found the same thing independently. It predates this branch — the empty-array spelling has always reached it:

```
{"A": {"main": [null]}, "B": {"main": [null]}}  → []   no error
{"A": {"main": [[]]},   "B": {"main": [[]]}}    → []   no error
{}                                              → "Multi-node workflow has no connections between nodes..."
```

Filed as #1101 with the measurement: tightening the scan newly reports 29 of the 2,352 bundled templates, and unlike the other corpus checks on this branch those are **not** uniformly false — workflow 3891 is a node-reference library whose showcase nodes genuinely are unwired. It needs its own pass, not a rider on this one.

czlonkowski:
Copilot's latest two reviews. Three genuinely new findings, all real, all reproduced against the built code before fixing — `c9bab5bf`.

**1. A rejected `operator.type` was interpolated raw, so a malformed operator threw instead of being reported.**

```
operator.type = {"toString": null, "valueOf": null}   (expressible in JSON)

before → validateOperatorStructure       THREW  "Cannot convert object to primitive value"
         validateConditionNodeStructure  THREW  (same — it calls the above)
after  → ["…operator: invalid type an object. Type must be a data type (…), not an operation name."]
```

Both `validate_workflow` and the create path call that function, so this surfaced as an opaque internal error — the exact defect class 2.84.1 through 2.84.3 closed elsewhere, left open in the operator path, and reachable only because #1097 points the check at the key real workflows use. A shared `describeOperatorValue` now names the value without coercing it, and the two `operation` interpolations in the sanitizer's `validateOperator` carry the same hazard, so they use it too.

**2. `validateNodeMetadata` guarded rule entries with a plain `typeof`, which arrays pass.**

```
rules.values: [[]]
before → ["Missing rules.values[0].conditions.options"]   (plus the accurate error from the condition validator)
after  → []
```

`sanitizeSwitchRule` and `EnhancedConfigValidator` both exclude arrays; this one didn't, so it added noise pointing at the wrong repair.

**3. `isOperationName` kept its own copy of the data-type list, so adding `any` never reached the sanitizer.**

```
operator = {type: "any"}      (no operation — genuinely invalid)

before → sanitized to {type: "string", operation: "any"}, validation then clean
after  → left as {type: "any"}, validation reports the missing operation
```

The rewrite invented an operation n8n has no such thing as, and because sanitization runs before validation it hid the real error on the way to the instance. It now reads `FILTER_OPERATOR_TYPES`.

Six new tests, each confirmed to fail against the pre-fix code. Full suite 7,371 green.

The other findings in the `01d783ed` review — the missing-`type` message and `ReplaceConnectionsOperation`, twice — were already fixed in `f82a1683`; Copilot re-anchors inline comments to current line numbers, so they reappeared against a commit that predated the fix.

Issue #620, opened 2026-03-07
feat(validator): Detect broken/malformed workflow connections


czlonkowski:
## Problem

The current `WorkflowValidator.validateConnections()` only checks for a known set of output keys (`main`, `error`, `ai_tool`) and silently ignores any other keys in the connections object. This means **malformed connections are never flagged**, even though they are effectively broken and invisible in the n8n UI.

### Real-world example

A workflow was generated with this connection structure:

```json
"Save to Google Sheets": {
  "1": [[{"node": "Format AI Integration Error", "type": "0", "index": 0}]],
  "main": [[{"node": "Webhook Success Response", "type": "main", "index": 0}]]
}
```

**Two bugs here:**
1. `"1"` is not a valid output key — it should be `main[1]`, not a top-level key
2. `"type": "0"` is invalid — should be `"type": "main"`

This connection exists in the JSON but **n8n doesn't render or execute it** — it's a ghost connection. The validator reported `0 invalid connections` and gave the workflow a passing grade on connections.

### Root cause

In `workflow-validator.ts`, `validateConnections()` only iterates over known keys:

```typescript
if (outputs.main) { ... }
if (outputs.error) { ... }
if (outputs.ai_tool) { ... }
```

Any unknown key (like `"1"`, `"0"`, `"output"`, etc.) is silently dropped.

---

### 1. Detect unknown output keys (P0 — easy win)

Flag any connection output key that isn't in the known set:
- `main` — standard outputs
- `error` — error outputs  
- `ai_tool`, `ai_languageModel`, `ai_memory`, `ai_outputParser`, `ai_embedding`, `ai_vectorStore`, `ai_retriever`, `ai_document`, `ai_textSplitter` — AI sub-node connections

Anything else should produce an error with a helpful message. Common LLM mistakes to detect:
- Numeric strings (`"0"`, `"1"`, `"2"`) — model confused output index with output type key
- Random names (`"output"`, `"success"`, `"error_output"`) — model invented a key

**Suggested error message:**
> Unknown connection output key "1" on node "Save to Google Sheets". Valid keys are: main, error, ai_tool, ai_languageModel, etc. If you meant to use the second output, use main[1] instead.

### 2. Validate `type` field in connection targets (P0 — easy win)

Every connection target has a `type` field that must be one of the known types (`main`, `error`, `ai_tool`, etc.). Currently we don't validate this at all. The example above has `"type": "0"` which is clearly wrong.

**Suggested error message:**
[...]

PR #1123, opened 2026-09-16
fix(validation): Switch branch counts, empty connection keys, validate_node operators, error-handler heuristic (#1100, #1101, #1103, #1111)

### LINKED ISSUE #1100: Switch branch-count validation reads the wrong rules key, and cannot be pointed at the right one as written
Split out of #1097, where the operator-structure half was fixed and this half was measured and deliberately left alone.

## The gap

`validateWorkflowStructure` (`src/services/n8n-validation.ts`) checks that a Switch node's output branches match its rules:

```ts
const rules = Array.isArray(params?.rules?.rules) ? params.rules.rules : [];
```

Same wrong key as #1097: n8n's Switch stores its rules under `parameters.rules.values` at typeVersion 3.2+, so this check has never run on a real Switch node. 319 of the 448 Switch nodes in the 2,352 bundled templates use `values`; the 5 that use `rules` are all typeVersion 1, which the `mode`/version handling puts out of scope.

## Why it cannot simply be pointed at `values`

Unlike the operator check, this one does not survive a false-positive pass. Simulating it over the corpus:

```
318 Switch nodes newly checked
→ 50 branch-count mismatches
→ 15 unconnected-output reports
→ 51 templates newly reporting errors
```

Close to all of them are false, in two distinct ways:

1. **`fallbackOutput` is not modelled** — 35 of the 50 mismatches have a `fallbackOutput` configured, which adds an output branch beyond the rules. 33 of those are exactly one branch over, which is the signature of this case.

   ```
   workflow 4072, node "Switch", typeVersion 3.2 → 3 branches, 2 rules, fallbackOutput set
   ```

2. **Trailing outputs are omitted by n8n** — the other 15 have *fewer* branches than rules, because n8n does not store trailing empty branches. A rule whose output goes nowhere is valid; it just routes items into nothing.

   ```
   workflow 4486, node "Switch", typeVersion 3.2 → 5 branches, 6 rules
   ```

## What a fix needs

- read `rules.values` as well as the legacy `rules.rules`
- account for `fallbackOutput` (and `options.fallbackOutput`) adding one output
- treat missing trailing branches as unconnected outputs rather than a count mismatch — i.e. the count check should be "branches ≤ rules (+1 for a fallback)", not equality
- re-run the corpus pass and land only if it comes back near zero

The same measurement for the operator check came back at 0 new errors across 944 rules, which is the bar.

### LINKED ISSUE #1101: A connection key with no targets vouches for its own node, so two isolated nodes escape the disconnected-node check
Found by Codex during the review gauntlet on #1099 and reproduced independently. Pre-existing; not caused by that PR, though #1096 adds a second shape that reaches it.

## The gap

The disconnected-node scan in `validateWorkflowStructure` (`src/services/n8n-validation.ts:402`) marks a source node connected before it has found a single edge:

```ts
Object.entries(workflow.connections).forEach(([sourceName, connection]) => {
  connectedNodes.add(sourceName); // Node has outgoing connection
  ...
```

The key's presence is treated as proof of an outgoing connection. A source key carrying no actual targets therefore vouches for its own node, and two nodes that vouch for each other escape the check entirely:

```
nodes: A, B (both n8n-nodes-base.noOp, nothing wired between them)

connections: {"A": {"main": [null]},  "B": {"main": [null]}}   → []   ← no error
connections: {"A": {"main": [[]]},    "B": {"main": [[]]}}     → []   ← no error
connections: {}                                                → "Multi-node workflow has no connections between nodes..."
```

So the *less* information the payload carries, the better the validation. An agent that emits an empty branch per node — a plausible shape for a draft — gets a clean bill of health for a workflow that does nothing.

`WorkflowValidator.flagOrphanedNodes` (`src/services/workflow-validator.ts:1699`) starts with the same line and suppresses its orphan warnings the same way.

## Relationship to #1096

The empty-array spelling has always reached this. The null spelling is new to callers as of v2.84.4, because the write schema used to reject a null branch before the connectivity scan ran. The defect itself is older than both.

## What a fix costs

Measured over the 2,352 bundled templates, comparing the current behaviour against a scan that only counts a source as connected once it has a real target:

```
templates newly reporting a disconnected node: 29 / 2352
```

Unlike the Switch branch-count case in #1100, these are **not** uniformly false. A sample:

| workflow | newly reported |
|---|---|
| 3891 "Explore n8n Nodes in a Visual Reference Library" | AI Agent, OpenAI, Basic LLM Chain, Information Extractor |
| 4013 "Automatic Jest Test Generation for GitHub PRs" | GH - Get PR, 2nd Node for Testing |
| 3862 "Automate Research-Based Newsletters with Perplexity" | Tavily WebSearch |
| 4058 "Homey Pro - Smarthouse integration with LLM" | xAI Grok Chat Model |

Workflow 3891 is a node-reference library whose showcase nodes genuinely are not wired to anything — a true positive. Others need checking one at a time before deciding whether the check or the template is wrong.

## Suggested approach

- Add `sourceName` to `connectedNodes` only after the source is observed to carry at least one target, in both scans.
- Walk the 29 templates and classify each newly-reported node as a real orphan or a check the scan should still forgive (AI sub-nodes reached only through `ai_*` connections are the shape most likely to need care).
- Land only if the residue is small and defensible, per the discipline the July validator audit set.

### LINKED ISSUE #1103: validate_node never checks IF/Switch operator structure, so it passes configs that validate_workflow and n8n_create_workflow reject
Spotted during the live testing on #1099 and confirmed to be broader than it first looked. Pre-existing; nothing in that PR caused it.

## The divergence

`validate_node` never runs operator-structure validation. `validateConditionNodeStructure` is called from `validateWorkflowStructure` (the `n8n_create_workflow` / `n8n_update_full_workflow` path) and from `WorkflowValidator` (`validate_workflow`), but not from `EnhancedConfigValidator`, which is what `validate_node` uses.

So every operator defect class passes `validate_node` and fails the other two. Measured against the built code, for both node types:

| node | operator | `validate_node` | `validate_workflow` |
|---|---|---|---|
| IF v2.2 | `{type: "string"}` — no `operation` | clean | error |
| IF v2.2 | `{operation: "equals"}` — no `type` | clean | error |
| IF v2.2 | `{type: "equals", operation: "equals"}` — operation name in the type field | clean | error |
| IF v2.2 | `"equals"` — a string | clean | error |
| IF v2.2 | `null` | clean | error |
| Switch v3.2 | all five of the same | clean | error |

Ten for ten. `EnhancedConfigValidator` is not unaware of these nodes — `validateSwitchNodeStructure` checks `rules.values` entries for `conditions` and `outputKey` — it just stops short of the operators inside them.

## Why it matters

This is the same shape as #1096, the divergence that prompted that PR: an agent validates, is told the config is fine, then the write is refused with no way to tell which validator to believe. The write path really does refuse — live-verified on n8n-test during #1099's testing, where `n8n_update_partial_workflow` and `n8n_create_workflow` both blocked a Switch carrying `operator: {"type": "any"}` with `rules.values[0].conditions.conditions[0].operator: missing required field "operation"`, while `validate_node` on the same config returned `valid: true`.

`validate_node` is the cheaper call and the one an agent reaches for while configuring a single node, so this is the validator most likely to be consulted before the mistake is made — and the one that stays quiet.

## Suggested approach

- Call `validateConditionNodeStructure` from `EnhancedConfigValidator` for `nodes-base.if`, `nodes-base.switch` and `nodes-base.filter`, mapping its strings onto the node-level error shape.
- Mind the version gates: the function already handles them (`typeVersion >= 2` for IF, `>= 3.2` plus rules-mode for Switch), but `validate_node` receives a config rather than a node, so the caller has to supply `typeVersion` — check what it has available before assuming.
- Run the usual false-positive pass over the bundled templates before landing. The equivalent sweep for the `rules.values` work in #1097 came back at 0 new errors across 944 rules, so the operator checks themselves are quiet on real data; the risk here is in the plumbing, not the predicate.

### LINKED ISSUE #1111: Error-output heuristic turns a Respond to Webhook fan-out into a hard error and prescribes a fix that breaks the success path
## Summary

`validate_workflow` raises a **hard error** (`valid: false`) whenever a node's `main[0]` fans out to more than one node and one of the targets *looks like* an error handler by **name** (`error`, `fail`, `catch`, `exception`) or **type** (`respondToWebhook`, `emailSend`). The suggested "fix" moves that node onto the error output and adds `onError: "continueErrorOutput"`. For the most common real case, a Respond to Webhook next to a side-effect node, following the fix breaks the workflow: the webhook would only ever answer when the upstream node **fails**.

## Reproduction (validate_workflow, profile `runtime`)

Webhook (`responseMode: responseNode`) → Set "Prepare Order" → [**Respond to Webhook**, Set "Save Order"]

```json
{"connections": {"Webhook": {"main": [[{"node": "Prepare Order", "type": "main", "index": 0}]]},
 "Prepare Order": {"main": [[{"node": "Respond to Webhook", "type": "main", "index": 0},
                             {"node": "Save Order", "type": "main", "index": 0}]]}}}
```

Result:

```
valid: false
Incorrect error output configuration. Nodes "Respond to Webhook" appear to be error handlers but are in main[0] (success output) along with other nodes.
... CORRECT (should be): main[1] = [Respond to Webhook] ... Also add: "onError": "continueErrorOutput" to the "Prepare Order" node.
```

The name heuristic fires the same way on ordinary fan-outs to nodes named e.g. "Log failed logins", "Catch-all router" or "Handle error (real error)".

## Where

[`src/services/workflow-validator.ts#L1154-L1195`](https://github.com/czlonkowski/n8n-mcp/blob/aef47e2f5f9dd7ac4ae0dd18cb9d2fddd58b3983/src/services/workflow-validator.ts#L1154-L1195): the `potentialErrorHandlers` filter uses `nodeName.includes('error' | 'fail' | 'catch' | 'exception')` and `nodeType.includes('respondtowebhook' | 'emailsend')`, then pushes to `result.errors`.

## Why it matters

Agents treat validator errors as must-fix. This one is a false positive on a correct, common topology, and its prescribed fix silently breaks the success path (the webhook request hangs until timeout). It was also hit during skill testing on nodes whose names merely contained "error".

## Suggested fix

- Don't raise an error from a name/type heuristic. At most a **warning**, and only when there's corroborating evidence (e.g. the source node already has `onError: continueErrorOutput` with nothing wired to `main[1]`, or the target reads `$json.error`).
- Drop `respondToWebhook` / `emailSend` from the type heuristic entirely. Both are normal success-path nodes.

Found while live-testing the n8n-skills pack (n8n 2.38.5, hosted staging on the latest release).

czlonkowski:
## Summary

- Fixes #1100. The write-path Switch branch-count check read only the legacy `rules.rules`, so it never ran on a current Switch. It now reads `rules.values` too, counts one extra output for `options.fallbackOutput: "extra"` and one for `onError: "continueErrorOutput"`, reports only more branches than outputs (n8n omits trailing branches), skips Switch v1 (four fixed outputs) and drops the unconnected-output error (13 published templates route a rule into nothing).
- Fixes #1101. A connection key with no targets (`main: [[]]`, `main: [null]`) no longer counts as a connection in either scan; triggers follow the same rule. The "no connections" check counts real targets too, while a malformed output value still lets its shape error speak.
- Fixes #1103. `validate_node` runs `validateConditionNodeStructure` for IF, Switch and Filter with the `@version` the handler merges in; a non-primitive `@version` no longer throws. `validate_workflow` keeps its direct call but skips messages the enhanced validator already reported, so every operator error appears exactly once (the enhanced validator keeps one error per property; a follow-up on that de-duplication lands with the native-Python PR).
- Fixes #1111. The error-handler name/type heuristic is no longer a hard error. The name is a hint appended to the existing warning for a node whose `continueErrorOutput` output is unwired; node type is not a signal.
- `validate_workflow` output counts: conditional nodes are resolved before dynamic metadata is rejected (Switch's bundled outputs are an expression, so its error-output warnings never fired), Switch counts rules plus fallback only when `extra`, Filter counts one output.

## Corpus (2,352 bundled templates)

| check | before → after |
|---|---|
| disconnected-node errors | 105 → 127 templates (22 new, all real orphans) |
| Switch unconnected-output errors | 0 → 0 (check removed) |
| Switch branch-count errors | 1 → 0 (the one was a v1 Switch) |
| "Incorrect error output configuration" hard errors | 46 → 0 |
| corroborated handler hints | 0 → 1 |
| false Filter "unmatched branch" warnings | 16 → 0 |

No other change apart from a random UUID in one duplicate-node-id suggestion.

## Review

Simplifier, Opus review, Codex pass. Adopted: the wired Switch error output and Switch v1 false positives (both blocking), duplicated operator errors, Filter's single output, Switch metadata being dynamic, the `@version` object crash, empty keys bypassing the "no connections" check, the self-loop suggestion, the always-paired warnings (merged into one), the unconditional fix text. Declined: promoting the orphan-node warning in `validate_workflow` to an error to mirror the write path (pre-existing severity split, out of scope).

## Verification

`npm run typecheck` clean; `npx vitest run tests/unit` 195 files, 6,579 passed; the MCP-protocol integration file for error-output validation passes.


czlonkowski:
Answers to the Copilot review (suppressed findings):

- Sticky note as suggested source: adopted; the suggestion now picks a connected executable node, then any other executable node, never a sticky note or the disconnected node itself. Test added.
- Switch v1 natural count: adopted; `getConditionalOutputInfo` returns 4 for v1 so the error output resolves at `main[4]` while the rule-count check still skips v1. Test added.
- Mode check in the output count: already in the previous commit; an expression-mode Switch counts `numberOutputs` and ignores a retained rule collection (Switch v3 has only `rules` and `expression` modes).
- Filter output count: declined. The bundled metadata for every Filter version (1 through 2.3) declares `outputs: ["main"]`; `outputNames` lists "Kept" and "Discarded" but n8n renders outputs from `outputs`. In the 2,352 bundled templates, 0 of 363 Filter v2+ nodes wire `main[1]`, which is what one would expect if a Discarded output were available in the editor. The cited integration test only shows that the Public API stores whatever connections it is given. Counting two outputs produced 16 false "no connections on the unmatched branch" warnings across the corpus, all removed by this change.

czlonkowski:
Answers to the third Copilot review:

- Numeric-string `@version`: adopted; it is stored as a number and `NaN`/`Infinity` fall back to version 1, with tests.
- Empty `rules.values`: adopted; the branch-count check keys on the collection being an array, so a Switch whose only output is the fallback is still counted. Test added.
- Non-rules, non-expression Switch mode: adopted; the output count is unknown for any other mode.
- Error-output slot without `onError` as a bounds error: declined. That is the #620 P1 decision, pinned by two tests (`Code` with `main[1]`, `IF` with `main[2]`); the "missing onError" warning is complementary, not a contradiction, since n8n drops items sent to an output the node does not have.

czlonkowski:
Answers to the fourth Copilot review, all adopted:

- A caller can no longer disable the operator checks by sending `@version: {}` or `undefined`: the `validate_node` handler keeps the node's database version unless the caller pins a finite number or numeric string.
- With one executable node and a sticky note the disconnected-node error no longer proposes a self-loop; the hint is omitted when there is no other executable node. Test added.
- A Switch with an empty rule collection and extra branches says it has no outputs instead of "indexed 0 to -1". Test added.

czlonkowski:
Answers to the fifth Copilot review:

- `any` missing from the operator hint: adopted.
- A connection object with an empty `node` counting as a connection: adopted; only a non-empty target string counts, with a test.
- "indices 0--1" for a Switch with no outputs: adopted; the bounds error says the node has no main outputs, with a test.
- Two conditions with the same defect collapsing to one in `validate_workflow`: declined as described. The messages carry the condition path (`conditions[0]`, `conditions[1]`), so they are distinct strings; the direct call skips only the exact message the config validator already reported and pushes the rest, which the test "reports two distinct malformed operators on the same IF node, each exactly once" pins. The property-level collapse inside the config validator is replaced by message-level de-duplication in #1125.

PR #1125, opened 2026-09-16
feat(validation): native-Python Code-node rules, pythonNative in validate_node, native Python guide (#1112, #1113, #1116)

### LINKED ISSUE #1112: validate_node reports "Code cannot be empty" for every Python Code node using language pythonNative
## Summary

`validate_node` reports **"Code cannot be empty"** (property `jsCode`) for every Code node that uses `language: "pythonNative"`, even when `pythonCode` has content. `pythonNative` is the only Python option n8n 2.x offers (`get_node` lists `javaScript` | `pythonNative`), so `validate_node` is effectively unusable for Python Code nodes. Workflow-level validation handles it correctly (it normalizes `pythonNative` → `python`).

## Reproduction

```js
validate_node({
  nodeType: "nodes-base.code",
  profile: "runtime",
  config: { language: "pythonNative", mode: "runOnceForEachItem",
            pythonCode: "row = _item[\"json\"]\nreturn {\"json\": row}" }
})
// → valid: false, errors: [{ property: "jsCode", message: "Code cannot be empty" }]
```

## Where

[`src/services/config-validator.ts#L603`](https://github.com/czlonkowski/n8n-mcp/blob/aef47e2f5f9dd7ac4ae0dd18cb9d2fddd58b3983/src/services/config-validator.ts#L603):

```ts
const codeField = config.language === 'python' ? 'pythonCode' : 'jsCode';
```

Compare [`node-specific-validators.ts#L1266`](https://github.com/czlonkowski/n8n-mcp/blob/aef47e2f5f9dd7ac4ae0dd18cb9d2fddd58b3983/src/services/node-specific-validators.ts#L1266), which already maps `pythonNative` → `python`.

## Suggested fix

Apply the same normalization in `ConfigValidator.validateCode` (`language === 'python' || language === 'pythonNative'`), and add a `validate_node` test with `pythonNative`.

### LINKED ISSUE #1113: Python Code-node validation still models Pyodide: legacy _input/_json/_now, blocked imports and classes pass; a valid single-dict return errors
## Summary

The Python Code-node checks still model the removed **Pyodide "Python (Beta)"** runtime. On n8n 2.x (`language: "pythonNative"`, task-runner native Python) several patterns that always fail at runtime pass validation, and one valid return shape is rejected. Everything below was verified by execution on n8n 2.38.5 with the `n8nio/runners` sidecar.

## Not flagged, but fail at runtime

| Code | Runtime result |
|---|---|
| `_input.all()`, `_input.first()`, `_json`, `_node["X"]`, `_now`, `_today`, `_jmespath(...)` | `NameError` (native Python exposes only `_items` in all-items mode and `_item` in each-item mode) |
| `_items` in each-item mode / `_item` in all-items mode | `NameError` |
| `item.json.field` | `AttributeError: 'dict' object has no attribute 'json'` |
| **any** `import` not on the instance allowlist, e.g. `import json` | `Security violations detected` / `Import of standard library module 'json' is disallowed. Allowed stdlib modules: none` (default allowlist is empty; n8n Cloud allows none). Only `requests`/`pandas`/`numpy`/`pip` are flagged today. |
| `class Foo: ...` | `__build_class__ not found` |
| `type(x)`, `getattr`, `hasattr`, `setattr`, `vars`, `dir`, `globals`, `locals`, `open`, `input`, `compile` | `NameError` (default `N8N_RUNNERS_BUILTINS_DENY`). Only `eval(`/`exec(` get a warning. |
| `x.__class__`, `"{0.__class__}".format(x)`, `__import__("json")` | `Security violations detected` (static rejection before execution) |
| `global counter` inside a nested function | `NameError` (user code runs inside a wrapper function; `nonlocal` works) |
| each-item mode returning a **list** | `A 'json' property isn't a dictionary [item 0]` |

Validating this with `validate_node` (after the `pythonNative` fix in #1112) or `validate_workflow` produces no Python-specific error:

```python
import json
all_items = _input.all()
total = sum(item.json.amount for item in all_items)
class Box:
    pass
return [{"json": {"total": total, "at": str(_now), "t": str(type(total)), "q": _jmespath(all_items[0]["json"], "name")}}]
```

## Flagged, but valid at runtime

- All-items mode `return {"count": len(_items)}` (a single dict, with or without a `json` key) → **"Return value must be a list of dicts"**. Native Python auto-wraps it into one item, and a list of plain dicts is auto-wrapped too. (The regex at [`#L1513-L1521`](https://github.com/czlonkowski/n8n-mcp/blob/aef47e2f5f9dd7ac4ae0dd18cb9d2fddd58b3983/src/services/node-specific-validators.ts#L1513-L1521) only matches when the code has no trailing newline, so the result also depends on formatting.)

## Misleading advice

- [`#L1877-L1891`](https://github.com/czlonkowski/n8n-mcp/blob/aef47e2f5f9dd7ac4ae0dd18cb9d2fddd58b3983/src/services/node-specific-validators.ts#L1877-L1891): Python `inputPatterns` are `['items', '_input']`, so each-item code that uses only `_item` gets "Code doesn't reference input data" with the suggestion "Access input with: items variable".
- [`#L2016`](https://github.com/czlonkowski/n8n-mcp/blob/aef47e2f5f9dd7ac4ae0dd18cb9d2fddd58b3983/src/services/node-specific-validators.ts#L2016): the JMESPath check looks for `_jmespath(` in Python, a function that no longer exists. It should flag it as unavailable instead.

## Suggested direction

A small native-Python rule set in [`validatePythonCode`](https://github.com/czlonkowski/n8n-mcp/blob/aef47e2f5f9dd7ac4ae0dd18cb9d2fddd58b3983/src/services/node-specific-validators.ts#L1386): error on the removed Pyodide globals and on dot access to `["json"]`; a warning on any `import` ("blocked unless the instance allowlists it; none on n8n Cloud"); errors on `class`, denied builtins and dunder attribute access; mode-aware `_items`/`_item` checks; accept single-dict returns in all-items mode and reject list returns in each-item mode.

Reference behaviour (verified): n8n-skills `skills/n8n-code-python/SKILL.md` (branch `docs/self-hosting-python-task-runners`).

### LINKED ISSUE #1116: python_code_node_guide documents the removed Pyodide runtime (_input.all(), _json, json module)
## Summary

`tools_documentation({topic: "python_code_node_guide"})` still documents the **Pyodide "Python (Beta)"** runtime that n8n 2.0 removed. Agents following it write code that fails on every n8n 2.x instance.

Current essentials output:

- "Access all items: `_input.all()`": raises `NameError` on native Python (only `_items` / `_item` exist)
- "Current item data: `_json`": raises `NameError`
- "Use json module for parsing" / "datetime for date handling": **every import is blocked by default** (`Security violations detected … Allowed stdlib modules: none`), and n8n Cloud allows none
- The example uses `all_items = _input.all()`

## Where

[`src/mcp/tools-documentation.ts#L469-L680`](https://github.com/czlonkowski/n8n-mcp/blob/aef47e2f5f9dd7ac4ae0dd18cb9d2fddd58b3983/src/mcp/tools-documentation.ts#L469) (`_input.all()` at L469, L481, L509, L623, L660).

## Suggested content (verified on n8n 2.38.5)

- `language: "pythonNative"`; only `_items` (all-items) / `_item` (each-item); dict access only
- No `_input`, `_json`, `_node`, `_now`, `_today`, `_jmespath`
- Imports blocked unless the instance allowlists them (self-hosted task-runner config); none on Cloud
- No `class`, `type()`, `getattr`/`hasattr`, or dunder access; `nonlocal` instead of `global`
- Return shapes: all-items accepts a list of `{"json": …}`, a list of plain dicts, or a single dict; each-item accepts a dict (`None` drops the item, a list errors)
- Self-hosted Docker needs the `n8nio/runners` sidecar, or every Python node fails with `Python runner unavailable: Python 3 is missing from this system`

A fuller reference: n8n-skills `skills/n8n-code-python/SKILL.md`.

czlonkowski:
## Summary

- Fixes #1112. `ConfigValidator.validateCode` normalizes `pythonNative` → `python`, so `validate_node` stops reporting "Code cannot be empty" on `jsCode`.
- Fixes #1113. `validatePythonCode` models n8n 2.x native Python per the verified skill contract: errors for the removed globals (`_input`, `_json`, `_node`, `_now`, `_today`, `_jmespath`) and the JavaScript habit `items`, `_items`/`_item` in the wrong mode, `.json` attribute access, `class`, denied builtins, dunder access (also inside format strings), `global` inside a function, list return in each-item mode; every import a `security`-typed warning (survives runtime/minimal). Single-dict return in all-items mode accepted. Pyodide-era checks removed from both layers. Names resolve per scope (def/lambda parameters, local defs, destructuring, annotated assignment); f-string fields are code; multi-line and single-line def headers handled; scans linear (200 KB adversarial shapes: 2–19 ms).
- Fixes #1116. `python_code_node_guide` rewritten for the native runtime (essentials and full); Python samples in task templates and `get_node` examples made native.
- Error de-duplication keys on property+type+message (was property+type, which collapsed six Python defects or five malformed IF operators into one). `missing_required` still collapses per property; two double-reporting sources suppressed at the source; a null required property no longer adds a type error.

## Corpus (2,352 bundled templates)

31 Python Code nodes in 14 templates: errors through the `validate_node` chain 15 → 40 (19 `_input`, 10 `.json` attribute, 5 `items`, 2 `getattr`, 2 `hasattr`, 2 `type()`/`__name__` in f-strings), warnings 81 → 107 (imports now visible under `runtime`); spot-checked, all true positives on Pyodide-era code. Non-Python delta: zero apart from one removed false enum error on a null optional value.

## Review

Simplifier, two Opus review rounds, two Codex passes, four fix rounds. Adopted: the de-duplication collapse (found by both reviewers), base-layer duplicates, profile filtering of the import warning, bound-name and scope handling, comprehension targets, parameter defaults, single-line and multi-line def headers, f-string fields and format-string dunders, import forms, indexed returns, trailing semicolons, quadratic helpers, null double-report, f-prefix case, dead dedup branch, docs qualifications, email example. Known limits (documented in code): a later binding in the same scope exempts an earlier reference; a query held in a variable is not followed.

Merge note: the `_jmespath` Code-node check removed in #1122 is replaced here by the removed-global error.

## Verification

`npm run typecheck` clean; `npx vitest run tests/unit` 196 files, 6,660 passed.


czlonkowski:
Answers to the Copilot review and the two CodeQL alerts (commit cf680e48):

- CodeQL `js/polynomial-redos` at the two return/import regexes: adopted; both are rewritten without adjacent overlapping quantifiers, and tab-padded 200 KB inputs are in the timing test (3 ms and 22 ms).
- Imported names as bindings, with `\` continuations and bracketed multi-line import lists: adopted; `from mod import _json` binds `_json`, and the per-module import warning reads the joined statement.
- Bare denied builtin references (`fn = eval`, `return type`): adopted; attribute access and keyword-argument names stay clean.
- Sandbox deny table in the guide: adopted; it now lists the complete set including `object`, `memoryview` and `breakpoint`.

czlonkowski:
Second Copilot review (commit b8084d10): 11 of 12 findings adopted — attribute access no longer matches the removed globals or mode variables, signed and leading-dot primitives are caught, one-line def suites belong to the def's scope, parenthesised returns are seen through, the format-string dunder check is limited to literals passed to `.format()`, comprehension and lambda bindings are local to their bracket group, bare `items(...)` errors while `d.items()` and `f(items=1)` stay clean, relative imports warn, the engine's `validateNodeOperation` runs through the enhanced validator, and the data-analysis template regains its stdev. Declined: the bundled `data/skills/n8n-code-python` skill, which is synced from the n8n-skills repository in a separate chore PR.

